PR #24, opened 2023-10-13
Fix #23: add resources directory and .yaml files.

### LINKED ISSUE #23: When requiring idpbuilder as a module, build fails due to `go:embed` pattern with invalid matches
Running into the error: ```pattern resources/*.yaml: no matching files found``` when using `go build`. 
Referenced line is in [/pkg/controllers/crd.go](https://github.com/cnoe-io/idpbuilder/blob/7b8a7cfbebda44206210bb435cbd755ab053dc3c/pkg/controllers/crd.go#L19).

From the [go embed documentation](https://pkg.go.dev/embed):
> Matches for empty directories are ignored. After that, each pattern in a //go:embed line must match at least one file or non-empty directory.
> If any patterns are invalid or have invalid matches, the build will fail.

When running `make` in idpbuilder, the `/pkg/controllers/resources` directory is generated. But when requiring idpbuilder in another project, the build fails without the generated directory.

hodgiwabi:


greghaynes:
I think youll need to remove this line too https://github.com/cnoe-io/idpbuilder/blob/main/.gitignore#L5

greghaynes (approved):

hodgiwabi:
`make` in the `dev` directory builds without errors when using the following in `go.mod`:
```
replace github.com/cnoe-io/idpbuilder => github.com/hodgiwabi/idpbuilder v0.0.0-20231013224858-faff040f02f8
```
where the fork branch includes [a similar `replace` directive](https://github.com/hodgiwabi/idpbuilder/blob/faff040f02f8f3be7a2f281398811a98c12e7baa/go.mod#L204) and contains [the `resources/*.yaml` files](https://github.com/hodgiwabi/idpbuilder/tree/faff040f02f8f3be7a2f281398811a98c12e7baa/pkg/controllers/resources).

greghaynes:
I think this test fail isnt you - pretty sure tests have been broken, merging.

Thanks @hodgiwabi

Issue #12, opened 2023-10-10
Backstage pod is crashing due to : Failed to pull image "backstage:1.0.0": rpc error: code = failed to pull and unpack image "docker.io/library/backstage:1.0.0"


cmoulliard:
## Issue

The Backstage pod created by `./idpbuilder create --buildName localdev` is crashing due to the following docker error:
```
Events:
  Type     Reason     Age                   From               Message
  ----     ------     ----                  ----               -------
  Normal   Scheduled  21m                   default-scheduler  Successfully assigned backstage/backstage-7bb974cbff-ffzj9 to localdev-worker
  Normal   Pulling    20m (x4 over 21m)     kubelet            Pulling image "backstage:1.0.0"
  Warning  Failed     20m (x4 over 21m)     kubelet            Failed to pull image "backstage:1.0.0": rpc error: code = Unknown desc = failed to pull and unpack image "docker.io/library/backstage:1.0.0"
: failed to resolve reference "docker.io/library/backstage:1.0.0": pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
  Warning  Failed     20m (x4 over 21m)     kubelet            Error: ErrImagePull
  Warning  Failed     19m (x6 over 21m)     kubelet            Error: ImagePullBackOff
  Normal   BackOff    6m36s (x64 over 21m)  kubelet            Back-off pulling image "backstage:1.0.0"
```

AFAIK, the official backtage image is now hosted on ghcr - https://github.com/backstage/backstage/pkgs/container/backstage. See what the helm chart references as image here: https://github.com/backstage/charts/blob/main/charts/backstage/values.yaml#L93-L99

```
  image:

    # -- Backstage image registry
    registry: ghcr.io

    # -- Backstage image repository
    repository: backstage/backstage

    # -- Backstage image tag (immutable tags are recommended)
    tag: latest
```

Warning: If we install backstage using ghcr.io repo, then we will get another error as a secret is missing
```
Events:
  Type     Reason     Age                 From               Message
  ----     ------     ----                ----               -------
  Normal   Scheduled  15m                 default-scheduler  Successfully assigned backstage/backstage-5c7f6948db-bqhnv to localdev-worker
  Normal   Pulling    15m                 kubelet            Pulling image "ghcr.io/backstage/backstage:latest"
[...]

PR #27, opened 2023-10-16
fix backstage manifests

nabuskey:
Updated the backstage manifests to:

1. Use in-memory database
2. Setup cluster role and binding to work with the default backstage image
3. Make the kind cluster available as a target cluster in Backstage

fixes: #12

### DIFF
--- a/pkg/apps/srv/backstage/install.yaml
+++ b/pkg/apps/srv/backstage/install.yaml
@@ -4,76 +4,64 @@ metadata:
   name: backstage
 ---
 apiVersion: v1
-kind: Secret
+kind: ServiceAccount
 metadata:
-  name: postgres-secrets
+  name: backstage
   namespace: backstage
-type: Opaque
-data:
-  POSTGRES_USER: YmFja3N0YWdl
-  POSTGRES_PASSWORD: aHVudGVyMg==
 ---
-apiVersion: v1
-kind: PersistentVolume
+# For invoking Argo Workflows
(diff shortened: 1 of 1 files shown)


pkg/apps/srv/backstage/install.yaml
```diff
@@ -90,15 +78,93 @@ spec:
...
+    auth:
+      environment: local
+      providers: {}
+    scaffolder: {}
+    catalog:
+      import:
+        entityFilename: catalog-info.yaml
+        pullRequestBranchName: backstage-integration
+      rules:
+        - allow: [Component, System, API, Resource, Location, Template]
```
blakeromano, line 154: is there a reason why are we not allowing domains into backstage?

jessesanford (approved):
LGTM

pkg/apps/srv/backstage/install.yaml
nabuskey: Good point, really no reason to prohibit it for local developments.

blakeromano: Although in general I could see us wanting to allow customization on things like `entityFilename` where some organizations use custom file names so I wonder if some of this being done through a GoLang template may make sense in the future.

nabuskey: Yea we can allow for more customization in the future. These manifests allow you to get started quickly locally.

nimakaviani (commented):
Shall we add ingress too, now that we are fixing it all?

PR #29, opened 2023-10-16
use localtest.me suffix to resolve localhost

jessesanford:


nimakaviani (changes requested):
There are a few more occurences of the same domain name that we should also fix

```
pkg/controllers/localbuild/controller.go
27:     gitServerIngressHostnameBase     string = ".idpbuilder.cnoe.io.local"

pkg/controllers/gitserver/controller.go
26:     gitServerIngressHostnameBase     string = ".idpbuilder.cnoe.io.local"

pkg/apps/srv/argocd/ingress.yaml
13:  - host: argocd.idpbuilder.cnoe.io.local
```

jessesanford:
> There are a few more occurences of the same domain name that we should also fix
> 
> ```
> pkg/controllers/localbuild/controller.go
> 27:     gitServerIngressHostnameBase     string = ".idpbuilder.cnoe.io.local"
> 
> pkg/controllers/gitserver/controller.go
> 26:     gitServerIngressHostnameBase     string = ".idpbuilder.cnoe.io.local"
> 
> pkg/apps/srv/argocd/ingress.yaml
> 13:  - host: argocd.idpbuilder.cnoe.io.local
> ```

These are used internally so are using the automatically generated service hostnames. We do need to figure out that duplication though. I will create a separate issue for that.

README.md
```diff
@@ -27,9 +27,9 @@ This can be useful in several ways:
 
 Kubernetes: `kubectl get pods`
 
-Argo: Visit https://argocd.idpbuilder.cnoe.io.local:8443/
+Argo: Visit https://argocd.idpbuilder.cnoe.io.localtest.me:8443/
 
-Backstage: http://backstage.idpbuilder.cnoe.io.local/
+Backstage: http://backstage.idpbuilder.cnoe.io.localtest.me/
```
nabuskey, line 32: I think it needs the port number here to connect to ingress-nginx

```
 http://backstage.idpbuilder.cnoe.io.localtest.me:8443
```

nimakaviani:
@jessesanford yeah I think the duplication is very confusing indeed. I think once we fix the README as @nabuskey pointed out, I am good with merging this for now.

cmoulliard:
Dont forget to change also the ingress host(s) if now we will use everywhere `*.idpbuilder.cnoe.io.localtest.me/` ?

nimakaviani (approved):
LGTM

README.md
cmoulliard: Yes as the we expose on the kind cluster and docker the ports 8443 and 8880. Question: Why do you export on kind/docker such ports instead to keep the standards: 80 and 443 ?

pkg/apps/srv/argocd/ingress.yaml
```diff
@@ -10,7 +10,7 @@ metadata:
 spec:
   ingressClassName: "nginx"
   rules:
-  - host: argocd.idpbuilder.cnoe.io.local
+  - host: argocd.idpbuilder.cnoe.io.localtest.me
```
cmoulliard, line 13: Does it make sense to use a so long DNS name `*.idpbuilder.cnoe.io.localtest.me` vs `*.idpbuilder.localtest.me` or even better `*.idp.localtest.me` ?

README.md
jessesanford: This is a good question. My guess is that they were using high number ports to avoid seeing to be a privileged user on some systems

cmoulliard: Such ports should be configurable and defaulted within the kind config file to: 80, 443. I will open a separate issue about that

Issue #5, opened 2023-09-28
Design pluggable "distro" packaging mechanism for building idp stack flavors


jessesanford:
We need to be able to allow for our different teams to build out their reference IDP stacks while not requiring all of them to be "hard coded" into the builder's "Embedded Apps". 

https://github.com/cnoe-io/idpbuilder/blob/56089e4ae3b27cf90641bfbff2a96c36dd5263e1/pkg/apps/resources.go#L20-L32

There are a number of ways this can be done. We should document a design proposal and alternatives / prior art. We should open it up for discussion in a PR and then ratify it with other interested orgs. We may or may not solution a straw man during discussion, but we should not merge to main until the design is agreed upon.

Some food for thought:

We should target the aws reference implementation install scripts as the first "flavor" of the idp distro since it is the most feature complete. See https://github.com/cnoe-io/reference-implementation-aws-user-friendly/blob/main/setups/backstage/install.sh

We can possibly make use of argo apps resource hooks to do imperative commands before the argo apps reconcile. See https://argo-cd.readthedocs.io/en/stable/user-guide/resource_hooks/

There are some issues with how the lifecycle hooks execute and when they are blocking vs not see the discussion here: https://github.com/argoproj/argo-cd/issues/9891

We want the idpbuidler to be a single binary with no dependencies other than docker.

We can possibly make use of Carvel imgpkg to bundle all of the apps and their resource definitions into a single image that can be used offline without a registry: https://carvel.dev/imgpkg/docs/v0.24.0/basic-workflow/#step-4-use-pulled-bundle-contents

Sealer is another project that has similar single binary like packaging of kubernetes and associated apps running on top of it. However it does seem to require the use of a registry? There may be some interesting prior art there especially in how it wraps k0s.
See: https://github.com/sealerio/sealer
See: https://www.alibabacloud.com/blog/sealer-a-kubernetes-based-distributed-application-package-%26-delivery-tool_599065

A research paper on kubernetes packaging systems: https://oparu.uni-ulm.de/xmlui/bitstream/handle/123456789/38625/promis20_01_baur.pdf?sequence=1

nabuskey:
I think we should focus on supporting plain yaml based approaches like helm and kustomize via GitOps. We could use a OCI image to bundle these files up, but that's just a way to get these files. 

[...]

PR #31, opened 2023-10-17
pluggable pakcages

nabuskey:
closes: #5


docs/pluggable-packages.md
```diff
@@ -0,0 +1,66 @@
...
+1. Support custom setup scripts and workflows that need to be executed in a defined order before a package becomes ready.
+
+## Proposal
+
+This document proposes the following:
+- Make ArgoCD and Argo Workflows hard requirements. 
+- Define packages as Argo CD Applications.
+- Imperative pipelines for configuring packages are handled with ArgoCD resource hooks and Argo Workflows.
+
+Currently, ArgoCD is a base requirement for both the AWS reference implementation and idpbuilder.
```
nimakaviani, line 35: The fact that we used Argo Workflows for the Data IDP will also make it fit nicely here.

docs/pluggable-packages.md
```diff
@@ -0,0 +1,66 @@
...
+- Make ArgoCD and Argo Workflows hard requirements. 
+- Define packages as Argo CD Applications.
+- Imperative pipelines for configuring packages are handled with ArgoCD resource hooks and Argo Workflows.
+
+Currently, ArgoCD is a base requirement for both the AWS reference implementation and idpbuilder. 
+
+
+## Future improvements
+
+- Extend support for using different tools for CD and imperative workflows. For example, we may consider supporting Tekton or Flagger for running imperative workflows.
```
nimakaviani, line 40: also worth calling out Flux for doing CD

docs/pluggable-packages.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Alternatives Considered
+
+#### Use OCI images as applications
+
+Projects such as Sealer and Kapp aim to use OCI images as the artifact to define and deploy multiple Kubernetes resources. This has a few advantages.
```
nimakaviani, line 46: Kapp doesnt need OCI images. Kapp + imgpkg allows to use OCI images.

docs/pluggable-packages.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Alternatives Considered
+
+#### Use OCI images as applications
+
+Projects such as Sealer and Kapp aim to use OCI images as the artifact to define and deploy multiple Kubernetes resources. This has a few advantages. 
+
+- Immutable single artifact that can be used to deploy to different clusters.
+- Simplicity. Application dependencies and supporting resources are defined and confined in the image.
+- Can use standard Kubernetes YAML files, Helm Charts, and Kustomize.
```
nimakaviani, line 50: also native support for versioning as well as signing and verification using industry standards.

docs/pluggable-packages.md
nimakaviani: We also need to provide details on where the Argo CD applications will come from. are we expecting the Argo CD apps to be pulled from remote git repositories or to utilize the embedded file system and git repository?

- if the latter, then adding support for remote repos will have to be discussed as a future plan.
- if the former, then we should discuss accessing the possibility to access git repositories as a dependency (with implications it might have on air-gapped environments etc)

docs/pluggable-packages.md
```diff
@@ -0,0 +1,124 @@
...
+1. No organization went through the same path to reach the point of needing a IDP.
+1. Past technology choices were made for specific needs of each organization. This results in organizationally unique technology inertia that may not be correctable in the near future.
+1. No industry consensus and standards for choosing specific IDP components.
+
+To fit the needs from different organizations, idpbuilder needs to be flexible in the what and how it can deploy different packages. Currently idpbuilder uses ArgoCD to install a [set of application](https://github.com/cnoe-io/idpbuilder/blob/56089e4ae3b27cf90641bfbff2a96c36dd5263e1/pkg/apps/resources.go#L20-L32), and they cannot be changed without modifying the source code.
+
+In addition, the git server uses Go's embed capability to serve contents to ArgoCD.  Because of this, to update package configurations in idpbuilder, it requires compiling the Go application. For Go developers this approach is straightforward to work with. For non-Go developers, this approach may be frustrating when needing to debug errors from an unfamiliar language while compiling the program.
```
cmoulliard, line 19: There is also another drawback to embed a component as a package and their resources as the resources deployed by ArgoCD from the local git server cannot be changed and will require to change the code -> recompile etc

docs/pluggable-packages.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Goals
+
+The proposal in this document should:
+
+1. Make the packages installed by `idpbuilder` configurable.
```
cmoulliard, line 25: I propose to also include as item:
- Define an easy way to declare the packages to be installed by the platform

docs/pluggable-packages.md
```diff
@@ -0,0 +1,124 @@
...
+1. Minimize the number of runtime dependencies necessary.
+1. Make it easy for end users to configure their packages.
```
cmoulliard, line 27: This sentence is similar from `Make the packages .... configurable`

docs/pluggable-packages.md
```diff
@@ -0,0 +1,124 @@
...
+1. Support custom setup scripts and workflows that need to be executed in a defined order before a package becomes ready.
+
+## Proposal
+
+This document proposes the following:
+- Make ArgoCD a hard requirement.
+- Define packages as Argo CD Applications (Helm, Kustomize, and raw manifests)
+- Imperative pipelines for configuring packages are handled with ArgoCD resource hooks.
```
cmoulliard, line 35: Is it the best approach to use Argocd Hooks to execute kubernetes jobs pre/post sync of resources vs using a real supply chain/pipeline tool as Tekton or Argo workflow ? A drawback to use hook is that they are executed for every resource sync and don't offer a kind a singleton hook which is only executed when a new application is installed.

cmoulliard: I would like also to mention that we should as much as we can use the argoCd pattern `App of Apps` - https://medium.com/dzerolabs/turbocharge-argocd-with-app-of-apps-pattern-and-kustomized-helm-ea4993190e7c
Using the [ApplicationSet](https://argo-cd.readthedocs.io/en/stable/user-guide/application-set/) and their generators will also help us to customize the Applications CR generated using by example the version of the pluggable package to be installed like its Helkm chart, etc

docs/pluggable-packages.md
```diff
@@ -0,0 +1,124 @@
...
+
+```yaml
+apiVersion: idpbuilder.cnoe.io/v1alpha1
+kind: Config
+metadata:
+  name: test
+spec:
+  packages:
+    - name: crossplane
+      preCreation:
```
cmoulliard, line 99: We need as additional fields:
- version
- description
- deployment_type: helm, kustomize, etc
- url: => external repository hosting the chart, oci image, raw resources, etc
- values: object including the values to be passed to the helm chart of yaml config file passed as parameter and mounted as volume

docs/pluggable-packages.md
```diff
@@ -0,0 +1,124 @@
...
+          secretRef:
+            name: job1
+            path: ./secret1
+          spec:
+            apiVersion: batch/v1
+            kind: Job
+            metadata:
+              name: pi
+      postCreation:
+        ...
```
cmoulliard, line 110: If we include jobs, how such a reconciliation will work as installing a package will require to of course let argoCd to install the resources within the target cluster but also that what we call jobs succeeded.

docs/pluggable-packages.md
```diff
@@ -0,0 +1,124 @@
...
+
+```
+
+This introduces a few problems. 
+
+1. Complexity.
+
+    This introduces a completely new mechanisms to manage applications. Current idpbuilder design is very simple with no concept of pipelining. It allows end users to define manifests at compile time, and apply them as Argo CD applications. Introducing this feature and maintaining it may involve significant time commitment.
```
cmoulliard, line 118: Having such a mechanism is definitively a plus as today creating a kubernetes platform is really a pain :-(
Even VMWare Tanzu TAP don't propose such a great solution top of their kapp + ytt + catalog solution

docs/pluggable-packages.md
nimakaviani: we can make Argo CD hooks run Argo workflows.

ideally actions taken as part of executing argo workflows will be no-ops in subsequent executions. but if that happens not to be the case, then maybe we consider triggering and running argo workflows separately. either way, I agree that Argo Workflows are better suited for doing imperative work.

docs/pluggable-packages.md
jessesanford: We can perform signing and verification of individual files like helm charts and kustomize, but not not the entire "package" which is very handy. Argo CD does allow us to require commits to be signed though, and that might be enough of an interim solution until we can get something that can work more like sigstore cosign. See https://argo-cd.readthedocs.io/en/release-1.8/user-guide/gpg-verification/

jessesanford:
Should we consider requiring signed commits for our reference "packages" as argo cd apps so that the idpbuilder can be built to trust only official reference packages? https://argo-cd.readthedocs.io/en/release-1.8/user-guide/gpg-verification/ we can also allow users to configure their own gpg keys to use for verification as flag to `idpbuilder create`? https://argo-cd.readthedocs.io/en/release-1.8/user-guide/gpg-verification/#configure-using-declarative-setup

docs/pluggable-packages.md
nabuskey: Yep Application Sets are a great idea to use with the in-cluster git server, especially for reference implementation stuff.

docs/pluggable-packages.md
nabuskey: I'll have to put more thoughts into this later, but my point is we should avoid creating another config format if possible.

nabuskey:
> Should we consider requiring signed commits for our reference "packages" as argo cd apps so that the idpbuilder can be built to trust only official reference packages? 

Did not know about this. We should 100% do this for our reference packages. 

My current thinking around git server stuff: 

1. I think embedded FS is quite useful for providing the reference implementation and possibly for use in air gapped environments. Great way to tie manifests to a idpbuilder release with no external dependencies or secrets to worry about.

2. As discussed in #32, Gitea is a great idea. This allows us to pull stuff from a remote repository using user credentials, push it to the in-cluster git server. We will need to handle kustomize remote references and helm subcharts to ensure everything we need is pushed to the in-cluster git server.

nabuskey:
Ok I've incorporated ideas posted above. 

TLDR: 
- Use gittea. 
- Support syncing local files to git
- Replace repo urls with local  repo urls in application spec. 
- No support for imperative pipelining spec until concrete use cases arise.

nabuskey:
About Git server contents:

Is mirroring content from remotes to in-cluster git server the right way to go? This becomes a bit more complicated once you start dealing with nested remotes. Especially for helm. We pretty much need to:

1. Get the root chart tgz
2. Extract, check dependencies. 
3. If there are subcharts, pull them. 
4. Repeat 1-3
5. Replace urls in dependencies for all subcharts. Tarball.
5. Push all to cluster. 

This breaks signing on charts as well, but not sure if it matters for local.

If we are to render them, things are a bit simpler. 
1. Get values.
2. Render. e.g. `helm template`
3. Push manifests to the cluster.
4. Update Argo App to use directory instead. 

This means we have to make sure we are using the same process to render manifests as ArgoCD does. So more libraries, dependencies, and potentially deal with rendering logic. This also pretty much won't allow you to take advantage of FluxCD source mechanisms IF we decide to support Flux in the future.

jessesanford:
> About Git server contents:
> 
> Is mirroring content from remotes to in-cluster git server the right way to go? This becomes a bit more complicated once you start dealing with nested remotes. Especially for helm. We pretty much need to:
> 
> 1. Get the root chart tgz
> 2. Extract, check dependencies.
> 3. If there are subcharts, pull them.
> 4. Repeat 1-3
> 5. Replace urls in dependencies for all subcharts. Tarball.
> 6. Push all to cluster.
> 
> This breaks signing on charts as well, but not sure if it matters for local.
> 
> If we are to render them, things are a bit simpler.
> 
> 1. Get values.
> 2. Render. e.g. `helm template`
> 3. Push manifests to the cluster.
> 4. Update Argo App to use directory instead.
> 
> This means we have to make sure we are using the same process to render manifests as ArgoCD does. So more libraries, dependencies, and potentially deal with rendering logic. This also pretty much won't allow you to take advantage of FluxCD source mechanisms IF we decide to support Flux in the future.
> 
> Also note that to support ArgoCD app of Apps, we need to render locally. Because app of apps could be inside a Helm Chart, Kustomize, or plain manfiests.
> 
> We can also pass local credentials to the cluster but I am not sure if that's a good practice. I can't be sure which credentials should be passed. Passing all would not be a good idea imo.

I don't see any drawback to rendering and then pushing to local gitserver. Am I missing something? Is it simply that we would have a dependency on helm as a lib in idpbuilder?

nabuskey:
> I don't see any drawback to rendering and then pushing to local gitserver. Am I missing something? Is it simply that we would have a dependency on helm as a lib in idpbuilder?

Yeah pretty much. Dependencies on Helm and Kustomize libraries. We'll probably have to get the rendering logic from ArgoCD so this will add more dependencies. I am leaning towards just rendering them instead. More dependencies but simpler.

jessesanford:
> @nabuskey Would it make sense only to mirror contents which are referenced by a local path? In the case of an argo app which already has a remote it might be ok to just use that remote directly. The main win for the gitserver IMO is removing the dependency on having to push packages to some external mirror, but if the package is already pushed this shouldnt be an issue.

I agree with this, I think the only issue is how much we can keep the code paths between local and remote as unified as possible. I would hate to have to handle two separate "pipelines" of artifact herding when there is a design choice that could result in only 1 being necessary. However, based on @nabuskey comment about needing to match the rendering to argo's means that the tradeoff complexity might be more of a burden. I will take a back seat on this decision as you guys know this plumbing better than me. My only concern is that we have a great local workflow for developers that doesn't require pushing to a remote just to test changes during quick iteration.

docs/pluggable-packages.md
```diff
@@ -0,0 +1,278 @@
...
+
+1. Make the packages installed by `idpbuilder` configurable.
+1. Minimize the number of runtime dependencies necessary.
+1. Define an easy way to declare the packages to be installed.
+1. Allow for fast local development feedback loop.
+
+## Proposal
+
+This document proposes the following:
+- Make ArgoCD a hard requirement.
```
greghaynes, line 33: I wonder if we could expand on this - I think we should make ArgoCD a hard requirement of our installers (at least for the time being), but we should also maintain separation between packages and CD tooling. e.g. I'd expect there to be separate backstage packages (helm/kustomize) from the Argo apps which install them. With this we leave the door open to supporting CD alternatives in the future. If this sounds reasonable could we include a line about "we will strive to maintain separation between packages and specific CD technologies (argo) to allow for future CD plug-ability.

greghaynes (approved):
One minor nit comments, LGTM

docs/pluggable-packages.md
nabuskey: Sounds reasonable to me. For this implementation, we will hard require ArgoCD but there is no reason this cannot be changed in the future.

docs/pluggable-packages.md
cmoulliard: "we can make Argo CD hooks run Argo workflows." 
There are some limitations using hooks = kubernetes jobs as ArgoCD delete them end like the pods of their execution. So, if users would like to view the logs, then this is impossible. If the job fails, then again it is also very difficult to access/vies the log of the newly pod created from a job as argocd kill/recreate them -every x seconds

cmoulliard:
To extend the proposition of @nabuskey to use internally a git server (= gitea), I would like to mention these points too:
- Make the git server accessible from admin users externally (UI or SSH)
- Clone ALL the resources within the git server (core components: ingress, argocd & cert manager + pluggable). Why ? To let an admin user to change a resource when problem took place during deployment and to avoid to rebuild `idpbuilder CLI`, etc
- Refactor the `localbuilds.idpbuilder.cnoe.io` CRD to better describe a package to be installed, source: URL, OCI image; type: core or pluggable (we should find a better name), packaging mode: helm, plain, kustomize, etc; config (= params)
- Use a Package or Component CRD to deploy everything on the platform including core and pluggable
- Convert the Package or Component CRD to their GitOps equivalent. Why: to be able to support different implementations: argocd, fluxcd, Crossplane, Terraform etc. So our CRD could be translated for ArgocD to an Application CRD, for FluxCD to a GitRepository, etc

nabuskey:
>     * Make the git server accessible from admin users externally (UI or SSH)
100% agreed. Being able to visualize and interact with git is valuable. 

>     * Clone ALL the resources within the git server (core components: ingress, argocd & cert manager + pluggable). Why ? To let an admin user to change a resource when problem took place during deployment and to avoid to rebuild `idpbuilder CLI`, etc

This was my intention as well. We will need to make ArgoCD manage itself but should be easy. I will clarify that in the doc. 

> 
>     * Refactor the `localbuilds.idpbuilder.cnoe.io` CRD to better describe a package to be installed.
>
>     * Component or Package CRD definition:
>       
>       * Name, labels
>       * Source: URL, OCI image, bucket
>       * Type: core or pluggable (we should find a better name)
>       * Packaging type: helm, plain, kustomize
>       * Config: = parameters
> 
>     * Use a Package or Component CRD to deploy everything on the platform including core and pluggable
> 
>     * Convert the Package or Component CRD to their GitOps equivalent. Why: to be able to support different implementations: argocd, fluxcd, Crossplane, Terraform etc. So our CRD could be translated for ArgocD to an Application CRD, for FluxCD to a GitRepository, etc


I think these make sense as this project matures. I have a few concerns at this stage where we want to focus on use cases in local settings. 

1.  We can provide abstraction for these implementations but there must be a well defined limit. These technologies vary widely between how they are used and how they work. We have to put a lot more thoughts on these alone. 
3. I also think this is getting a bit close to application modeling. Application modeling is hard. There has been attempts at modeling applications like OAM with Kubevela. As far as I am aware, there are no solutions that the industry is converging to. We need to be careful abut introducing another.

4. Dependencies and overheads. We will have to import a lot of dependencies and match rendering logic versions. 

These are great discussion points to follow up after this PR. I think we should keep it simple and nail the local experience first. I think it's a low hanging fruit that we can immediately see its values on.

Issue #22, opened 2023-10-13
Backstage component is not deployed as service and no ingress route is available


cmoulliard:
## Issue

The backstage component is not deployed as a service and no ingress route is published too
```
k get ingress -A
NAMESPACE          NAME                    CLASS   HOSTS                               ADDRESS     PORTS   AGE
argocd             argocd-server-ingress   nginx   argocd.idpbuilder.cnoe.io.local     localhost   80      10m
idpbuilder-local   gitserver-embedded      nginx   embedded.idpbuilder.cnoe.io.local   localhost   80      10m
```

cmoulliard:
I included the following service and ingress yaml to the existing backstage yaml 
```
apiVersion: v1
kind: Service
metadata:
  name: backstage
  namespace: backstage
spec:
  ports:
  - name: http
    port: 80
    protocol: TCP
    targetPort: 7007
  - name: https
    port: 443
    protocol: TCP
    targetPort: 7007
  selector:
    app: backstage
  sessionAffinity: None
  type: ClusterIP
---
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: backstage
  namespace: backstage
  annotations:
    nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
    nginx.ingress.kubernetes.io/ssl-passthrough: "true"
spec:
  ingressClassName: nginx
  rules:
    - host: backstage.idpbuilder.cnoe.io.local
      http:
        paths:
          - backend:
              service:
                name: backstage
                port:
                  name: https
            path: /
            pathType: Prefix
```
but cannot access the URL: 
```
curl -k http://backstage.idpbuilder.cnoe.io.local:8880/
<html>
<head><title>308 Permanent Redirect</title></head>
<body>
<center><h1>308 Permanent Redirect</h1></center>
<hr><center>nginx</center>
</body>
</html>

or

curl -k https://backstage.idpbuilder.cnoe.io.local:8443/
curl: (35) LibreSSL/3.3.6: error:1404B42E:SSL routines:ST_CONNECT:tlsv1 alert protocol version
```

cmoulliard:
I fixed this issue here: https://github.com/cnoe-io/backstage-app/issues/13

PR #33, opened 2023-10-17
Add service and ingress resources

cmoulliard:
- Issue: #22 
- Adding the definition about the Service and Ingress kubernetes resource to expose backstage using the host: `https://backstage.idpbuilder.cnoe.io.local:8443`

Remarks: 
- The UI reports as error `Fail to load entity kinds` when we navigate to `https://backstage.idpbuilder.cnoe.io.local:8443/catalog`. This problem is most probably due to cors badly defined as described here: https://github.com/backstage/backstage/issues/15403 
- If later, we also deploy cert manager, then we could add an annotation to the ingress resource to get the TLS secret populated by the cert manager

### DIFF
--- a/pkg/apps/srv/backstage/install.yaml
+++ b/pkg/apps/srv/backstage/install.yaml
@@ -130,7 +130,7 @@ metadata:
 spec:
   ingressClassName: nginx
   rules:
-    - host: backstage.idpbuilder.cnoe.io.local
+    - host: backstage.idpbuilder.cnoe.io.localtest.me
       http:
         paths:
           - backend:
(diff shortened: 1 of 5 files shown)


pkg/apps/srv/backstage/install.yaml
```diff
@@ -116,17 +152,17 @@ data:
   app-config.yaml: |
     app:
       title: CNOE
-      baseUrl: http://localhost:3000
+      baseUrl: "https://backstage.idpbuilder.cnoe.io.local:8443"
```
nabuskey, line 155: These needs to be either `local.me` or `local.cnoe.io`, depending on which one we use in #29

cmoulliard: I suggest to change all the URLs (including what we have here + documentation when a PR will be created to fix issue 29

greghaynes: Yep - we're going to hopefully merge https://github.com/cnoe-io/idpbuilder/pull/29 asap which does this, we should change it here assuming we merge that one first.

cmoulliard: I updated the DNS name to use cnoe.io.localtest.me @greghaynes

cmoulliard:
As you merged my commits in another PR, I will then close this PR @jessesanford Next time, we should keep my PR and not creating another one ;-)

jessesanford:
> As you merged my commits in another PR, I will then close this PR @jessesanford Next time, we should keep my PR and not creating another one ;-)

Hey sorry for merging your work without you online. I was creating the same code changes to fix backstage. Unfortunately I could not rebase your fork sorry for pulling the branch here in order to do so, all of your commits and the credit was attributed to you. In fact the full PR was taken in its complete state in order to retain your intent. If the PR was not ready for merge just mark it as WIP or draft or "don't merge". However I tested it and it was nearly exactly the code that I had prepared to commit.

PR #38, opened 2023-10-18
use the right reference when deleting the image

### LINKED ISSUE #36: Image digest is used for image ID
The image ID field is filled with image digest, not image ID. 

https://github.com/cnoe-io/idpbuilder/blob/1a9acbb3d71ac797dae14bbb32533507b30538e4/pkg/controllers/gitserver/image.go#L53-L62

`PushImage` returns:
https://github.com/cnoe-io/idpbuilder/blob/1a9acbb3d71ac797dae14bbb32533507b30538e4/pkg/apps/image.go#L55-L63

Because digest is passed to Docker client, the test fails. https://github.com/cnoe-io/idpbuilder/blob/1a9acbb3d71ac797dae14bbb32533507b30538e4/pkg/controllers/gitserver/image_test.go#L55

Did we meant to return the digest instead of id in this? https://github.com/cnoe-io/idpbuilder/blob/1a9acbb3d71ac797dae14bbb32533507b30538e4/pkg/apps/image.go#L23

nimakaviani:


nabuskey (commented):
This fixes the test but my issue was around returning image digests vs id in the status field. 

We can include both but they need to be clearly labeled.

nimakaviani:
lets track that separately. I will create another issue for it.

nabuskey:
I think the issue describe the problem fine. Let's unlink the issue from this PR and keep it open.

nimakaviani:
agreed. that's why I didnt create a new one. lets keep that around but get the test fixed for now.

nabuskey (approved):
LGTM

Does not resolve Issue #36  but it does fix this particular test

Issue #32, opened 2023-10-17
[suggestion] Use gitea as embedded git server


cmoulliard:
## Suggestion

In order to simplify the development, maintenance of idpbuilder, I would like to suggest to use [gitea](https://docs.gitea.com/) as embedded git server.

Why ?
- Project is maintained by a huge community
- Gitea can be easily installed on a k8s cluster using a helm chart
- It offers a Web UI similar to github
- We can ssh using the local port 22 mapped to a nodePort on the cluster (e.g. 32222)
- The web +UI can be exposed top of an ingress ropute: gitea.idpbuilder.localtest.me

## HowTo

To play locally with gitea:

1. Create a kind cluster
```
curl -s -L "https://raw.githubusercontent.com/snowdrop/k8s-infra/main/kind/kind.sh" | bash -s install --delete-kind-cluster --port-map 32222:22
```
2. Deploy the gitea helm chart
```
cat <<EOF > helm-values.yml
redis-cluster:
  enabled: false
postgresql:
  enabled: false
postgresql-ha:
  enabled: false

persistence:
  enabled: false

gitea:
  admin:
    username: "gitea_admin"
    password: "gitea_admin"
    email: "gi@tea.com"
  config:
    database:
      DB_TYPE: sqlite3
    session:
      PROVIDER: memory
    cache:
      ADAPTER: memory
    queue:
      TYPE: level

service:
  ssh:
    type: NodePort
    nodePort: 32222
    externalTrafficPolicy: Local

ingress:
  enabled: true
  className: nginx
  hosts:
    - host: gitea.localtest.me
      paths:
        - path: /
          pathType: Prefix
EOF
helm install gitea gitea-charts/gitea -n gitea --create-namespace -f helm-values.yml
```
3.  Using the UI, create a `dummy` git repo and import your ssh pub key
```
user/password to be used to logon `gitea_admin/gitea_admin` on http://gitea.localtest.me
Repo to be created: "http://gitea.localtest.me/gitea_admin/dummy"
Import your public key "http://gitea.localtest.me/user/settings"
```

4. Create a local git repo + files -> commit
5. Add the remote git repo
```
git remote add origin git@localhost:gitea_admin/dummy.git"
```

6. `git push -u origin main`

:-)

@greghaynes

greghaynes:
Love the idea!

greghaynes:
[...]

PR #39, opened 2023-10-18
install gitea

nimakaviani:
fixes 32


cmoulliard:
Why do you install gitea as kubernetes resources and not using a Helm chart instead as I described in my issue #32 ?

pkg/controllers/localbuild/resources/gitea/values.yaml
cmoulliard, line 1: This values file is only needed if you install gitea as a helm chart (and this is what I recommend to do) as it allows to externalize the configuration/parameters. 

If you deploy it using the helm chart, then additional parameters will be needed to expose the gitea server as UI on the cluster, etc as documented here #32

nimakaviani:
The values file was used in `helm template` to generate and put the vanilla k8s resources under `resources/gitea/k8s`. I kept it there as I am still experimenting with what resources we want to configure (including the ingress and service information you mentioned in #32). we can remove it once we finalize the PR,

If we go down the route of installing gitea using the helm chart, we will have to either pull the helm go libraries in or create a dependency to helm, which I'd like to avoid. 

As we discussed in #5 handling of any package or tool outside the initial bootstrapping should be handed over to Argo Apps, which can then take care of helm or kustomize or any other package types. That's why I am hesitant to introduce helm as part of the bootstrapping mechanism. its a draft PR, so open to suggestions here.

cmoulliard:
> If we go down the route of installing gitea using the helm chart, we will have to either pull the helm go libraries in or create a dependency to helm, which I'd like to avoid.

Why ? You are also today creating a hard dependency today as you must maintain the ArgoCD YAML file able to install argocd as core component + controller tracking its installation while using helm go lib we could more easily install the core components (argo cd, vault, argo workflow, cert manager) before to delegate to ArgoCD the job to install the `pluggable` Application CRs = Components

cmoulliard:
> we can remove it once we finalize the PR,

Ok.

nabuskey:
@nimakaviani Is this going to be ready for review soon? I would like to get Gittea integration going.

nabuskey:
We should reference the secret for the admin user credentials instead of directly using it in the pod template. See: https://gitea.com/gitea/helm-chart#admin-user

This way other controllers can reference it and get the credentials to talk to Gitea api. This does mean we have to helm template then add another manifest for the secret, which is a bit annoying and may cause confusions among users. So we should document and warn.

nabuskey:
For helm values, we will also need to specify endpoint domain and URLs. Otherwise, API responses from Gitea for anything referencing base url has the default values of `git.example.com`. 

For my local testing, I am using this, then port-forward. 

```yaml
gitea:
  admin:
    existingSecret: gitea-admin-secret
  config:
    database:
      DB_TYPE: sqlite3
    session:
      PROVIDER: memory
    cache:
      ADAPTER: memory
    queue:
      TYPE: level
    server:
      DOMAIN: localtest.me
      ROOT_URL: 'http://localtest.me:3000'
```

jessesanford (approved):
Love it. Thanks for drying up those local installs.

pkg/controllers/localbuild/argo.go
```diff
@@ -5,105 +5,47 @@ import (
...
-	}
-
-	return k8s.ConvertRawResourcesToObjects(scheme, rawResources)
-}
-
-func newArgoNamespace() *corev1.Namespace {
-	return &corev1.Namespace{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: "argocd",
+func (r LocalbuildReconciler) ReconcileArgo(ctx context.Context, req ctrl.Request, resource *v1alpha1.Localbuild) (ctrl.Result, error) {
```
nabuskey, line 19: Let's use pointer receiver. I know we don't need pointers for these but rest of them do. We are currently mixing value and pointer receivers.

pkg/controllers/localbuild/nginx.go
```diff
@@ -3,133 +3,38 @@ package localbuild
...
-
-func newNginxNamespace() *corev1.Namespace {
-	return &corev1.Namespace{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: nginxNamespace,
-		},
-	}
-}
 
 func (r LocalbuildReconciler) ReconcileNginx(ctx context.Context, req ctrl.Request, resource *v1alpha1.Localbuild) (ctrl.Result, error) {
```
nabuskey, line 19: Same thing here. Let's use pointer receiver.

pkg/controllers/localbuild/gitea.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+const (
+	giteaNamespace = "gitea"
+)
+
+//go:embed resources/gitea/k8s/*
+var installGiteaFS embed.FS
+
+func (r LocalbuildReconciler) ReconcileGitea(ctx context.Context, req ctrl.Request, resource *v1alpha1.Localbuild) (ctrl.Result, error) {
```
nabuskey, line 19: Same thing here. Let's use pointer receiver.

pkg/controllers/localbuild/nginx.go
```diff
@@ -3,133 +3,38 @@ package localbuild
...
-	nginxNamespace  string = "ingress-nginx"
-	nginxDeployment string = "ingress-nginx-controller"
+	nginxNamespace string = "ingress-nginx"
 )
 
 //go:embed resources/nginx/k8s/*
 var installNginxFS embed.FS
-var timeout = time.After(3 * time.Minute)
-
-func RawNginxInstallResources() ([][]byte, error) {
```
nabuskey, line 32: Can we keep the ability to get embedded resources outside of these packages? Keep the functions or export the variable. The git repository reconciler needs a way to retrieve these to sync with gittea repos. You don't have to pass local dir ref and we have a distinction between repositories created with embedded contents vs local directory. 

```yaml
apiVersion: idpbuilder.cnoe.io/v1alpha1
kind: GitRepository
metadata:
  name: argocd
  namespace: default
spec:
  giteaURL: "http://localhost:3000"
  source:
    embeddedAppName: "argocd"
    type: embedded
```

api/v1alpha1/localbuild_types.go
```diff
@@ -20,7 +20,14 @@ type EmbeddedArgoApplicationsPackageConfigSpec struct {
 	Enabled bool `json:"enabled,omitempty"`
 }
 
+// GitConfigSpec controls what git server to use for the idpbuilder
+// It can take on the values of either gitea or gitserver
+type GitConfigSpec struct {
+	Type string `json:"type,omitempty"`
```
nabuskey, line 26: Wish there was a way to put the reference to secrets that contain gitea admin credentials in Spec or Status. I think for this implementation we can just document secret requirements. Not a blocker imo, but we need to open an issue.

pkg/controllers/localbuild/installer.go
```diff
@@ -0,0 +1,182 @@
...
+type EmbeddedInstallation struct {
+	name         string
+	resourcePath string
+	namespace    string
+
+	// skips waiting on expected resources to become ready
+	skipReadinessCheck bool
+
+	// name and gvk pair for resources that need to be monitored
+	expectedResources map[string]schema.GroupVersionKind
```
nabuskey, line 37: I think the name of this field should be `monitoredResources`. I first thought this was for making sure resources specified are present during installation.

api/v1alpha1/localbuild_types.go
nimakaviani: well the secret name is hardcoded here, so maybe we can just use it for now. i like the idea https://github.com/cnoe-io/idpbuilder/pull/39/files#diff-6eb0197d5d12e1ca681eac36170032f5d01e997ea21bdbd87803a302f9df254fR582-R590

pkg/controllers/localbuild/argo.go
nimakaviani: good call. not sure how it fell through

pkg/controllers/localbuild/nginx.go
nimakaviani: lets revisit it as part of pushing the files to the repos in the next PR. We will make adjustments where needed.

Issue #28, opened 2023-10-16
Fix tests


nabuskey:
Currently tests are broken. Need to fix them.

An example failed run
https://github.com/cnoe-io/idpbuilder/actions/runs/6532866095/job/17750266601

#36

cmoulliard:
The problem is related to this error `registry_test.go:31: Error reconciling registry: Error response from daemon: network kind not found`

cmoulliard:
Controller_test fails as content compared diverges from the existing kind config
```
/opt/homebrew/opt/go/libexec/bin/go tool test2json -t /Users/cmoullia/Library/Caches/JetBrains/IntelliJIdea2023.2/tmp/GoLand/___cluster_test_go.test -test.v -test.paniconexit0 -test.run ^\QTestGetConfig\E$
=== RUN   TestGetConfig
    cluster_test.go:55: Got config: # two node (one workers) cluster config
        # Kind kubernetes release images https://github.com/kubernetes-sigs/kind/releases
        kind: Cluster
        apiVersion: kind.x-k8s.io/v1alpha4
        containerdConfigPatches:
        - |-
          [plugins."io.containerd.grpc.v1.cri".registry.mirrors."localhost:5001"]
            endpoint = ["http://idpbuilder-testcase-registry:5000"]
        nodes:
        - role: control-plane
          image: "kindest/node:v1.26.3"
          kubeadmConfigPatches:
          - |
            kind: InitConfiguration
            nodeRegistration:
              kubeletExtraArgs:
                system-reserved: memory=4Gi
                node-labels: "ingress-ready=true"
          extraPortMappings:
          - containerPort: 80
            hostPort: 8880
            protocol: TCP
          - containerPort: 443
            hostPort: 8443
            protocol: TCP
          -
        - role: worker
          image: "kindest/node:v1.26.3"
          kubeadmConfigPatches:
          - |
            kind: JoinConfiguration
            nodeRegistration:
              kubeletExtraArgs:
                system-reserved: memory=4Gi
```

cmoulliard:
Do you plan to create a PR to fix the issues with the test ? 
I created a commit that you could already include. See: cc922d4

nabuskey:
Yes, I am documenting issues with tests. I think there are multiple issues with the tests see: 

#36 
#37 

As for your fix, I would love to see a PR :) @cmoulliard

cmoulliard:
[...]

PR #45, opened 2023-10-18
PR to fix the issues reported by the Issue 28

cmoulliard:
- PR to fix the issues reported by the Issue #28


cmoulliard:
kind cluster config comparison is failing part of the cluster_test.go test as additional chars are included such as: `\"`, '\t',  `  ` are added - see here: https://gist.github.com/cmoulliard/53fb4cfb0c40c3041469b70477f1562c#file-gistfile1-txt-L35-L82. I can reproduce the issue locally but I don't know if this is a go issue with diff, OS, etc

cmoulliard:
Proposition to use testcontainer: #49 
WDYT ? @nabuskey

cmoulliard:
Is a docker registry created OOTB when we execute `make test` part of a github workflow ? 
The `kind` network should be added to /etc/hosts or registered as docker network otherwise `registry_test.go:31: Error reconciling registry: Error response from daemon: network kind not found` will continue to fail 
@nabuskey

nabuskey:
> Is a docker registry created OOTB when we execute `make test` part of a github workflow ?

Yeah docker is one of tools that's included in the Ubuntu image that GitHub provides. 

I think I want to hold off on moving to test containers, if we use Gitea like you suggested, we don't need to build images.

cmoulliard:
> I think I want to hold off on moving to test containers, if we use Gitea like you suggested, we don't need to build images.

I agree with you. Nevertheless, if it is need to bump a container during some tests execution, using TestContainer will simplify our life

cmoulliard:
As most of the issues have been fixed, I propose to close this PR. WDYT ? @nabuskey

nabuskey:
Yeah let's close this and thanks for working on this.  I do like your approach with TestContainers though. If a need arises for something like this, we should definitely consider it.

Issue #37, opened 2023-10-18
Git server controller test does not include built-in controllers


nabuskey:
env test is set up here
https://github.com/cnoe-io/idpbuilder/blob/fd507f50b4733388b156aa3b7ae499daab91901d/pkg/controllers/gitserver/controller_test.go#L31

`UseExistingCluster` is set to false (zero value)

https://github.com/kubernetes-sigs/controller-runtime/blob/7db81c7b27468b517d70ce3f7da49b134ec51fec/pkg/envtest/server.go#L152

Per this doc, no built-in controllers run https://book.kubebuilder.io/reference/envtest.html#testing-considerations
So no deployment is created and the test fails. 

Not sure if there is a easy way around this.

cmoulliard:
So, if I understand correctly the problem, it is then needed that:
- We install a kind cluster and local registry first
- Next we deploy the needed controllers 
before to execute the tests: `make test` or `run a test file using ide tool` ?
@nabuskey 

Were the test cases working before ?

nabuskey:
Not sure if it was working before. @greghaynes ? 

To me looks like it was written before writing the code being tested, which is good. It's just that the `envtest` didn't provide a good explanation in code. 
We could use kind cluster to test them all but not sure if it's necessary. The advantage of `envtest` is the lightweight nature to test your controller fast. To me, waiting for `gotResource.Status.DeploymentAvailable` may not be necessary. We are testing our controller logic, not the deployment controller. We could look at the deployment object and ensure it's what we expect.

cmoulliard:
> We could use kind cluster to test them all but not sure if it's necessary.

Correct. We just need to run as testing environment kubebuilder where the idp controllers are registered for the tests.
There is nevertheless one issue as the image_test.go requires an existing container registry where image is pushed/removed. Do we have to perform such a test ? Does it make sense - not sure

PR #53, opened 2023-10-20
fix git controller test

nabuskey:
fixes: #37 

Built-in controllers don't run so we can just look at objects instead of waiting for a deployment to be ready.


nimakaviani (approved):
LGTM

Issue #50, opened 2023-10-20
Setup Go Releaser


nabuskey:
We should setup a Release GitHub action. This should be invoked when a new tag in the form `v*.*.*` is pushed. 

Something like:

```
name: go-releaser
on:
  push:
    tags:
      - 'v[0-9]+.[0-9]+.[0-9]+'

permissions:
  contents: write

jobs:
  goreleaser:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3
        with:
          fetch-depth: 0
      - run: git fetch --force --tags
      - uses: actions/setup-go@v4
        with:
          go-version: '>=1.20.0'
      - uses: goreleaser/goreleaser-action@v5
        with:
          distribution: goreleaser
          version: '~> 1.20.0'
          args: release --clean
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

PR #57, opened 2023-10-23
add go releaser

nabuskey:
Added a release GH action. Not supporting Windows for now because I have tested in Linux and MacOS only. 

Example Action: https://github.com/nabuskey/idpbuilder/actions/runs/6619626640

Release created: https://github.com/nabuskey/idpbuilder/releases/tag/v0.0.1

closes: #50

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,32 @@
+name: release
+on:
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+
+permissions:
+  contents: write
+
+jobs:
+  release:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+        with:
+          fetch-depth: 0
+      - run: git fetch --force --tags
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,32 @@
...
+        with:
+          fetch-depth: 0
+      - run: git fetch --force --tags
+      - uses: actions/setup-go@93397bea11091df50f3d7e59dc26a7711a8bcfbe # v4.1.0
+        with:
+          go-version: '1.20.0'
+      # Ensure generation tools run
+      - name: build
+        run: |
+          OUT_FILE=/tmp/localidp make build
```
nimakaviani, line 24: lets put it back to `idpbuilder` for now

.goreleaser.yaml
```diff
@@ -0,0 +1,33 @@
+project_name: localidp
```
nimakaviani, line 1: same here

```suggestion
project_name: idpbilder
```

nimakaviani: and in other places please

nimakaviani (changes requested):

nimakaviani (approved):
LGTM

PR #66, opened 2023-10-25
fix argocd controller tests

nabuskey:
```bash
➜  cnoe/idpbuilder git:(fix/tests) ✗ yq ea '[.] | length' pkg/controllers/localbuild/resources/argo/install.yaml
55
➜  cnoe/idpbuilder git:(fix/tests) ✗ ls -lh pkg/controllers/localbuild/resources/argo/
total 2152
-rw-r--r--  1 mccloman  staff   1.0M Sep 29 10:06 install.yaml
```

So the number of resources should be 1 and the number of objects in the argocd resource file should be 55. 

We will need to make a bit more robust tests in the future but I want to fix these tests so that they don't show up as failures every PR.


nimakaviani (approved):
LGTM

PR #67, opened 2023-10-26
create kind network

nabuskey:
The test assume we have a docker network named "kind". Since the test only cares about registry image, we'll just create the network prior to reconciling. 

Note that we will not need this functionality going forward. If we did, I think the way @cmoulliard did with test containers is the way to go. I want to fix tests so we don't run into test failures for unrelated PRs. 

```bash
ubuntu@ip-10-192-11-119:~/idpbuilder$ go test github.com/cnoe-io/idpbuilder/pkg/kind -run '^\QTestReconcileRegistry\E$'
ok  	github.com/cnoe-io/idpbuilder/pkg/kind	4.495s
ubuntu@ip-10-192-11-119:~/idpbuilder$ docker images
REPOSITORY   TAG       IMAGE ID   CREATED   SIZE
ubuntu@ip-10-192-11-119:~/idpbuilder$ docker network ls
NETWORK ID     NAME      DRIVER    SCOPE
906d8469762b   bridge    bridge    local
90ea394ecb86   host      host      local
44350ca9e837   none      null      local
```

fixes: #28

### DIFF
--- a/pkg/controllers/gitserver/image_test.go
+++ b/pkg/controllers/gitserver/image_test.go
@@ -3,6 +3,11 @@ package gitserver
 import (
 	"context"
 	"fmt"
+	"github.com/cnoe-io/idpbuilder/pkg/kind"
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/go-connections/nat"
+	"io"
+	"os"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


nabuskey:
Well, it passes the tests now. I will say that this is very fragile. 

Good news is that we don't need these tests going forward.

pkg/controllers/gitserver/image_test.go
```diff
@@ -39,23 +47,74 @@ func TestReconcileGitServerImage(t *testing.T) {
...
+	for {
+		if time.Now().After(endTime) {
+			t.Fatalf("Timed out waiting for port %d to be available", kind.ExposedRegistryPort)
+		}
+		conn, cErr := net.DialTimeout("tcp", net.JoinHostPort("0.0.0.0", strconv.Itoa(int(kind.ExposedRegistryPort))), time.Second*3)
+		if cErr != nil {
+			break
+		}
+		conn.Close()
+		time.Sleep(waitInterval)
```
nimakaviani, line 76: cosmetics here but pushing the check to a `go func` would be nicer

```suggestion
        ready := make(chan bool)
        endTime := time.After(waitTimeout)
	go func() {
		for {
		   ....
		   if cErr == nil {
		       close(ready)
		   }
		}
	}()

	select {
	case <-timeout:
		err := errors.New("Timeout")
		log.Error(err, "Didn't reconcile Gitea on time.")
		return ctrl.Result{}, err
	case <-ready:
		log.Info("Gitea is ready!")
		resource.Status.GiteaAvailable = true
	}
```

pkg/kind/registry_test.go
```diff
@@ -19,23 +20,44 @@ func TestReconcileRegistry(t *testing.T) {
...
-	if err != nil {
-		t.Fatalf("Error reconciling registry: %v", err)
+	defer dockerCli.ContainerRemove(ctx, cluster.getRegistryContainerName(), types.ContainerRemoveOptions{Force: true})
+	waitTimeout := time.Second * 90
+	waitInterval := time.Second * 3
+	endTime := time.Now().Add(waitTimeout)
+
+	for {
+		if time.Now().After(endTime) {
+			t.Fatalf("Timed out waiting for registry. recent error: %v", err)
```
nimakaviani, line 43: same as above

nimakaviani (approved):
LGTM

yay, first green build. thanks!

pkg/controllers/gitserver/image_test.go
nabuskey: I'm using the same format that was used here: https://github.com/cnoe-io/idpbuilder/blob/f6a3d8b5aebc77eae0b72d4ed24343f9174db0a8/pkg/controllers/gitserver/controller_test.go#L120

If we are using timeout, we should use the same logic everywhere or make it reusable. not worth the time effort imo.

PR #69, opened 2023-10-26
start registry container

nabuskey:
Fixes the issue with the test on git image where you get a error message like [this](https://github.com/cnoe-io/idpbuilder/actions/runs/6647695202/job/18063537873#step:4:73)

```
--- FAIL: TestReconcileGitServerImage (19.02s)
    image_test.go:44: reconcile error: Error response from daemon: Get "http://localhost:5001/v2/": dial tcp [::1]:5001: connect: connection refused
    image_test.go:48: Invalid or no Image ID in status: ""
    image_test.go:59: Removing docker image: Error response from daemon: invalid reference format
FAIL
```

It starts a registry container so it can get the digest. Definitely not the best way to go about testing these but we shouldn't need the code tested going forward.

### DIFF
--- a/pkg/controllers/gitserver/image_test.go
+++ b/pkg/controllers/gitserver/image_test.go
@@ -3,6 +3,11 @@ package gitserver
 import (
 	"context"
 	"fmt"
+	"github.com/cnoe-io/idpbuilder/pkg/kind"
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/go-connections/nat"
+	"io"
+	"os"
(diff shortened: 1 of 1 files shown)


nimakaviani (approved):
Nice!

I saw this failure when I was testing the fix for creating the kind network. lets get it over the line and get to green before cutting the first release

nabuskey:
Closing this because build issues are closely related to https://github.com/cnoe-io/idpbuilder/pull/67

Will handle both in that PR

PR #70, opened 2023-10-27
run go vet

nabuskey:
We need to make sure go vet is run before merge.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,9 +30,9 @@ vet: ## Run go vet against code.
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
 ifeq ($(RUN),)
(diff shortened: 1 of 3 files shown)


nabuskey:
The only way I can get the tests to run successfully is by running them one by one because there is a hard requirement for these tests to have a container running with the same port. I can't guarantee which one runs first. I also can't guarantee container run requests will succeed even with wait conditions because they could both send requests at the same time. 

I see four ways: 
1. Rework the controllers to use docker interfaces instead so we can use our implementation in tests. 
2. Use test containers. 
3. Make the tests run with one by one like it is done currently in this PR. this is bad because it affects all tests and slow.
4. Remove one of tests.

nimakaviani (approved):
i'd say it is fine to require the sequence for v0.1

we will do some major refactor in the next release

PR #79, opened 2023-11-10
Support creating gitserver from local path

greghaynes:


cmoulliard:
Why are creating an image ? @greghaynes

greghaynes:
@cmoulliard This is what we do internally so I wanted to add it in case its valuable before the pluggable packages work is implemented. One we have gitea this whole workflow should go away.

nabuskey:
Closing this since we have #100 open now.

PR #82, opened 2023-11-14
Implement git repository controller

### LINKED ISSUE #59: API type for Git repository
The new type is used to reconcile git repositories. 

Given a ArgoCD application spec like this:

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
spec:
  sources:
    - path: argo-workflows
      repoURL: "files://test-package"
      targetRevision: HEAD
```

idpbuilder should:
1. Create a new repository in the in-cluster git server. 
2. Populate the repository with the contents from the local `test-package` directory. 
3. Update the `repoURL` field(s) to the URL from step 1. 
4. Set the source type to 'directory'
5. Any changes to this directory should be synced during reconciliation. 

API type may look something like

```go
type GitRepositorySpec struct {
	Source GitRepositorySource `json:"source,omitempty"`
}

type GitServerSource struct {
	Directory string `json:"directory"`
        // +kubebuilder:validation:Enum:=dir,remote
        Type string `json:"type"`
        // +kubebuilder:validation:Enum:=helm,kustomize,manifests
        // +kubebuilder:default:=manifests
        Format string `json:"format"`
}

```

### LINKED ISSUE #60: Git Repository reconciler
Create a reconciler for https://github.com/cnoe-io/idpbuilder/issues/59

Target the manifest format only.

nabuskey:
This implements the git repository types and controller. 

Syncs the contents of a local directory to an in-cluster gitea repository for ArgoCD to pick up. 

We need https://github.com/cnoe-io/idpbuilder/pull/39 merged first.

closes #59, closes #60


pkg/controllers/gitrepository/controller.go
```diff
@@ -0,0 +1,280 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	DefaultBranchName = "main"
+	// DefaultGiteaUsername and DefaultGiteaPassword taken from gitea helm chart. https://gitea.com/gitea/helm-chart#gitea
```
nabuskey, line 30: We need a way to get these values dynamically. If the gitea object can expose that information, we can do so. Otherwise we will have to rely on a secret naming convention. e.g. we will always look for a secret named this in this namespace.

pkg/controllers/gitrepository/controller.go
```diff
@@ -0,0 +1,280 @@
...
+	DefaultGiteaUsername = "gitea_admin"
+	DefaultGiteaPassword = "r8sA8CPHD9!bt6d"
```
nimakaviani, line 32: I exposed the secret name for gitea under `resource.Status.GiteaSecretName` for the local build. let's use that instead. The value is based on what we have pushed to the registry here.

Also if we want to make the username and password fancier, we can certainly do it. I dont see particularly a need though. Gitea is set to only accept local connections anyway.

https://github.com/cnoe-io/idpbuilder/blob/11a9914536cbd8ee8da9f5d79e8924f1e595ec0c/pkg/controllers/localbuild/resources/gitea/k8s/install.yaml#L583-L590C4

pkg/controllers/gitrepository/controller.go
```diff
@@ -0,0 +1,280 @@
...
+
+type RepositoryReconciler struct {
+	client.Client
+	GiteaClientFunc GiteaClientFunc
+	Recorder        record.EventRecorder
+	Scheme          *runtime.Scheme
+}
+
+func getRepositoryName(repo v1alpha1.GitRepository) string {
+	return fmt.Sprintf("%s-%s", repo.Name, repo.Namespace)
```
nimakaviani, line 52: shouldnt it be `Namspace-Name` instead. easier to reason about it.

pkg/controllers/gitrepository/controller.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+func getOrganizationName(repo v1alpha1.GitRepository) string {
+	return fmt.Sprintf("%s-%s", globals.ProjectName, repo.Namespace)
+}
+
+func getCredentials(name, namespace string) (string, string, error) {
+	return DefaultGiteaUsername, DefaultGiteaPassword, nil
+}
```
nimakaviani, line 61: i like the forward thinking here.

api/v1alpha1/gitrepository_types.go
```diff
@@ -0,0 +1,66 @@
...
+	// +kubebuilder:validation:Optional
+	EmbeddedAppName string `json:"embeddedAppName"`
+	// Path is the absolute path to directory that contains Kustomize structure or raw manifests.
+	// This is required when Type is set to local.
+	// +kubebuilder:validation:Optional
+	Path string `json:"path"`
+	// Type is the source type.
+	// +kubebuilder:validation:Enum:=local;embedded
+	// +kubebuilder:default:=local
+	Type string `json:"type"`
```
nimakaviani, line 25: shouldnt default be embedded?

pkg/controllers/gitrepository/controller.go
```diff
@@ -0,0 +1,280 @@
...
+	requeueTime          = time.Second * 30
+	gitCommitAuthorName  = "git-reconciler"
+	gitCommitAuthorEmail = "invalid@cnoe.io"
+)
+
+type GiteaClientFunc func(url string, options ...gitea.ClientOption) (GiteaClient, error)
+
+func NewGiteaClient(url string, options ...gitea.ClientOption) (GiteaClient, error) {
+	return gitea.NewClient(url, options...)
+}
```
nimakaviani, line 42: not used?

nimakaviani (commented):
Looks neat, thanks. left a few comments.

Also, looks like the bit that actually hooks things up and runs this in `RunControllers` is not here in this commit. intentional?

nimakaviani:
also looks like you've got some conflicts to resolve

pkg/controllers/gitrepository/controller.go
```diff
@@ -0,0 +1,280 @@
...
+	return true
+}
+
+func getEmbedded(name string) ([][]byte, error) {
+	switch name {
+	case "argocd":
+		return localbuild.GetRawInstallResources()
+	case "nginx":
+		return localbuild.RawNginxInstallResources()
+	default:
```
nimakaviani, line 254: ah, now I see your point! this breaks since in the previous commit I removed these functions. maybe we put the functions back where needed?

nimakaviani: Also, probably makes more sense to decouple resource creation from the actual git repository controller code and place it all here where we reconcile other resources as well

https://github.com/cnoe-io/idpbuilder/blob/11a9914536cbd8ee8da9f5d79e8924f1e595ec0c/pkg/controllers/localbuild/controller.go#L280-L286C2

jessesanford: I agree with this, let's have one place in the sources for the creation of resources no matter what their install type

api/v1alpha1/gitrepository_types.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type GitRepositorySpec struct {
+	Source GitRepositorySource `json:"source,omitempty"`
+	// GiteaURL is the base URL of GitTea server
+	// +kubebuilder:validation:Required
+	GiteaURL string `json:"giteaURL"`
```
greghaynes, line 11: Can we not infer this? IIUC we have this statically set here: https://github.com/cnoe-io/idpbuilder/blob/main/pkg/controllers/localbuild/resources/gitea/values.yaml#L40, so could we either statically set that or use a shared field which templates this out?

My concern is that, as a user, why should I need to know the gitea url and supply it?

api/v1alpha1/gitrepository_types.go
```diff
@@ -0,0 +1,66 @@
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type GitRepositorySpec struct {
+	Source GitRepositorySource `json:"source,omitempty"`
```
greghaynes, line 8: How do we see the lifecycle of this resource? One thing that bugs me about the current implementation is this coupling of out of cluster state with resources that live inside the cluster and it looks like we're going to have to answer a similar question here:

* If the CLI is stopped, should this resource still exist?
* If so, then if the CLI is restarted should it resume syncing this gitrepository?
* If we do that, do we need to perform some guarding so multiple git repositories arent fighting to write to the same destination?
* What if someone runs multiple instances of the CLI?

api/v1alpha1/gitrepository_types.go
nabuskey: That url will always resolve as `127.0.0.1`. I thought we wanted to make this usable for external clusters as well in the future? 

As a user, you don't have to supply this. Creation of GitReposiroty CRs should be handled in the CLI.

api/v1alpha1/gitrepository_types.go
nabuskey: I think these questions apply to all resources we create so let's open an issue and talk about it. I agree we need to handle created resources better because currently we don't handle lifecycles of our resources at all. We don't even provide a way to delete created clusters.

pkg/controllers/gitrepository/controller.go
nabuskey: Yep that makes sense.

pkg/controllers/gitrepository/controller.go
nabuskey: This needs to be used. It's just an implementation of the controller itself.

api/v1alpha1/gitrepository_types.go
nimakaviani: +1 on what Manabu suggested. this should be an optional field that resolves to localhost if empty.

Also more generically it should say `GitURL`

greghaynes: A default sounds great!

nabuskey:
This PR's become a bit too large for my liking but it implements the entire workflow with Gitea. 

All emebdded manifests are pushed to gitea server under the admin user organization. We may want to make this configurable in the future. 

image


Once they are pushed, argo applications are created. They reference the repos created above. From here, ArgoCD takes care of the rest. The Git repository contents are reconciled once for embedded applications. If users wants to reconfigure things, they can do so in the Gitea UI.


image

Once [this PR ](https://github.com/cnoe-io/idpbuilder/pull/91)merges, I just need to rebase and build should pass.

pkg/controllers/localbuild/gitea.go
```diff
@@ -3,21 +3,32 @@ package localbuild
...
 )
 
 const (
-	giteaNamespace = "gitea"
-	// hardcoded secret name from what we have in the yaml installation file
+	// hardcoded values from what we have in the yaml installation file.
+	giteaNamespace   = "gitea"
 	giteaAdminSecret = "gitea-admin-secret"
+	// this is the URL accessible outside cluster. resolves to localhost
+	giteaIngressURL = "http://gitea.cnoe.localtest.me:8880"
```
nimakaviani, line 19: `8443` or `8880` for the port number?

nimakaviani (approved):
few minor updates. otherwise looks good.

One higher level question, for the backstage embeded app, shall we use the helm chart with its image set to our CNOE backstage image and let Argo CD handle it, rather than having the raw resources?

pkg/controllers/localbuild/controller.go
```diff
@@ -278,9 +279,123 @@ func (r *LocalbuildReconciler) SetupWithManager(mgr ctrl.Manager) error {
 }
 
 func (r *LocalbuildReconciler) ReconcileArgoAppsWithGitea(ctx context.Context, req ctrl.Request, resource *v1alpha1.Localbuild) (ctrl.Result, error) {
-	log := log.FromContext(ctx)
+	logger := log.FromContext(ctx)
```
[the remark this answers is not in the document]
nabuskey: Problem with using `log` is that it collides with package import like this https://github.com/cnoe-io/idpbuilder/blob/09608f07787012836c6d57eda4b691eefb8ed283/pkg/controllers/localbuild/controller.go#L21 

I agree that we should use logger instead of log across the project. I'll file an issue.

pkg/controllers/localbuild/gitea.go
cmoulliard: > Should be 8880. We should consider using TLS in the future, but it's all http for gitea as of now.

I don t think that we should append additional ports to the URL as we do now:
- `https://argocd.<host_IP>.<DOMAIN>:8443`, 
- `https://gitea.<host_IP>.<DOMAIN>:8880`, 
- `https://backstage.<host_IP>.<DOMAIN>:8443`,
- etc

Why: ?
- Adding more and more new ports will make the work of the users not really cool and easy as they will have to remember the ports to be used for each component
- Only default HTTP(S) ports are needed and will be recognized by the browser: 80, 443 automatically. 
- As the ingress controller which is acting as a [reverse proxy](https://en.wikipedia.org/wiki/Nginx) using nginx is able to forward from URL: `https://backend>.<host_ip>.<domain>` the packets to the target backends: argocd, backstage, vault, gitea, etc then adding new ports make the job to configure ingress + kind part of the idpbuilder hardder instead of having Ingress YAML files

api/v1alpha1/gitrepository_types.go
```diff
@@ -0,0 +1,75 @@
...
+	// +kubebuilder:validation:Optional
+	EmbeddedAppName string `json:"embeddedAppName"`
+	// Path is the absolute path to directory that contains Kustomize structure or raw manifests.
+	// This is required when Type is set to local.
+	// +kubebuilder:validation:Optional
+	Path string `json:"path"`
+	// Type is the source type.
+	// +kubebuilder:validation:Enum:=local;embedded
+	// +kubebuilder:default:=embedded
+	Type string `json:"type"`
```
cmoulliard, line 29: Can you describe what is the purpose of the type and what will happen if we select one the values: `local`, `embedded`, etc  ?

api/v1alpha1/gitrepository_types.go
```diff
@@ -0,0 +1,75 @@
...
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Pattern=`^https?:\/\/.+$`
+	GitURL string `json:"gitURL"`
+	// SecretRef is the reference to secret that contain Git server credentials
+	// +kubebuilder:validation:Optional
+	SecretRef SecretReference `json:"secretRef"`
+}
+
+type GitRepositorySource struct {
+	// +kubebuilder:validation:Enum:=argocd;backstage;crossplane;gitea;nginx
```
cmoulliard, line 19: crossplane should be removed from the enum and we should keep what currently we support or point to a link = doc page enumerating them

api/v1alpha1/gitrepository_types.go
```diff
@@ -0,0 +1,75 @@
...
+type GitRepositoryStatus struct {
+	// LatestCommit is the most recent commit known to the controller
+	// +kubebuilder:validation:Optional
+	LatestCommit Commit `json:"commit"`
+	// ExternalGitRepositoryUrl is the url for the in-cluster repository accessible from local machine.
+	// +kubebuilder:validation:Optional
+	ExternalGitRepositoryUrl string `json:"externalGitRepositoryUrl"`
+	// GitRepositoryUrl is the url for the in-cluster repository accessible within the cluster.
+	// +kubebuilder:validation:Optional
+	GitRepositoryUrl string `json:"gitRepositoryUrl"`
```
cmoulliard, line 52: Why don't you name it InternalGitRepositoryUrl as it is used for internal purposes ? If we use gitea which is accessible by the users, does that makes sense that we keep ext vs internal repos ?

pkg/controllers/localbuild/gitea.go
nabuskey: I agree from the user friendliness standpoint. The reason we are not using default http and https ports is that allocation of lower number port (< 1024) requires higher privileges that some organizations do not allow end users to have. I would like to use default ports but I would rather have the assurance that port allocation will work for all users. Maybe we can make it configurable in the future.

api/v1alpha1/gitrepository_types.go
nabuskey: Local indicates the gitea repository contents will come from a local directory. E.g. `/home/ubuntu/myrepo`

Embedded indicates the repository contents will come from what is embedded in the go binary.

api/v1alpha1/gitrepository_types.go
nabuskey: We will probably have to keep it since the reference implementation uses it. For now, I am keeping it because the purpose of this PR is to move the git functionality over from embedded git server to gitea.

api/v1alpha1/gitrepository_types.go
nabuskey: Good idea I will rename the field.

We need to distinction because the external and internal git repo url (`gitea.cnoe.localtest.me`) because the domain name will always resolve to 127.0.0.1. Therefore this URL is not usable by ArgoCD to fetch manifests from the gitea server.

PR #83, opened 2023-11-15
install nginx out of band

nimakaviani:
- remove embedded nginx installation
- add a new reconciler to deploy nginx resources
- use the same resources previously installed via argo
- wait for it to successfully deploy before proceeding
- fix minor issues


nimakaviani:
lining up the PRs to reduce review overload. This is required for the gitea server's ingress to be installed out of band and for the domain names to become available to the gitea client

pkg/k8s/client.go
```diff
@@ -36,5 +36,8 @@ func EnsureObject(ctx context.Context, kubeClient client.Client, obj client.Obje
 	if err != nil {
 		return err
 	}
+
+	// hacky way to restore the GVK for the object after create corrupts it. didnt dig. not sure why?
```
nabuskey, line 40: So calling create mess with gvk somehow? Very strange.

pkg/controllers/localbuild/nginx.go
```diff
@@ -0,0 +1,134 @@
+package localbuild
```
nabuskey, line 1: We need tests for these.

pkg/controllers/localbuild/nginx.go
```diff
@@ -0,0 +1,134 @@
...
+)
+
+const (
+	nginxNamespace  string = "ingress-nginx"
+	nginxDeployment string = "ingress-nginx-controller"
+)
+
+//go:embed resources/nginx/k8s/*
+var installNginxFS embed.FS
+var timeout = time.After(5 * time.Minute)
```
nabuskey, line 30: 5 minutes seems a bit too much? Users may end up needing to wait 5 minutes just to see something went wrong. 
Maybe one minute?

nabuskey (changes requested):

pkg/controllers/localbuild/nginx.go
```diff
@@ -0,0 +1,134 @@
...
+					}
+				}
+			}
+
+			time.Sleep(1 * time.Second)
+		}
+	}(installObjs)
+
+	select {
+	case <-timeout:
```
jessesanford, line 119: I am not well versed enough to know if this is the idiomatic way of waiting for the resource to become available. Isn't it also possible to requeue like so: https://github.com/cnoe-io/idpbuilder/blob/1440d25e8d74ba44fdc31a2818fa0efaf4e6b8a8/pkg/controllers/localbuild/controller.go#L167

jessesanford (approved):
Just some style points. Take them as you will.

pkg/k8s/client.go
nimakaviani: yeah. this literally wasted 2 3 hours of my time the other day. very strange. will come back to this once I unblock the gitea work.

pkg/controllers/localbuild/nginx.go
nimakaviani: darn, you noticed!

nabuskey: We can punt tests for now imo. We should work on getting gitea working first. 
Just need to open an issue and it can be a good first time issue for new contributors.

pkg/k8s/client.go
greghaynes: I wonder if we're applying some kind of object which hasnt had its gvk registered in our scheme?

nimakaviani: no, the object has its GVK up to the point we make the call to `create` it. it comes back from that call and it doesnt have the GVK in there anymore.

pkg/controllers/localbuild/nginx.go
nimakaviani: it could take more than one minute for the ingress controller to become ready. i'll bring it down though to maybe 2 or 3 mins and see how that works

nabuskey: Sounds good.

pkg/controllers/localbuild/nginx.go
nimakaviani: interesting. didnt know about this. let me look into it.

nimakaviani: ok yeah so I did dig and `RequeueAfter` only schedules to rerun the reconciliation loop. The check we have in the code is to make sure `Nginx.Status.AvailableReplicas > 0` so we can have the localbuild status for Nginx readiness set to true.

nabuskey (approved):
LGTM

PR #88, opened 2023-11-20
correct endpoints and use correct api version

nabuskey:
generated with 

```bash
helm template  my-gitea gitea-charts/gitea --version 9.5.1  -f pkg/controllers/localbuild/resources/gitea/values.yaml > pkg/controllers/localbuild/resources/gitea/k8s/install.yaml && echo "---
apiVersion: v1
kind: Secret
metadata:
  name: gitea-admin-secret
type: Opaque
stringData:
  username: giteaAdmin
  password: giteaPassword\n" >> pkg/controllers/localbuild/resources/gitea/k8s/install.yaml 
```


nimakaviani (approved):
LGTM

Issue #94, opened 2023-11-22
[Bug]: Ensure error handling in build


nabuskey:
### What is your environment, configuration, and command?

The error returned here is ignored and we could end up in a limbo. 

https://github.com/cnoe-io/idpbuilder/blob/458852cae89657eea8f54d94b156f46e61f5397c/pkg/build/build.go#L149

Also, the localbuilder resources are cluster scoped. Specifying the default namespace ends up not triggering updates because runtime object and what we expect do not match. We should not specify namespace for this resource type. 

https://github.com/kubernetes-sigs/controller-runtime/blob/2154ffbc22e26ffd8c3b713927f0df2fa40841f2/pkg/controller/controllerutil/controllerutil.go#L397

If CreateOrUpdate call fails, we should let the users know and exit.

PR #96, opened 2023-11-24
handle localbuild creation error and look for cluster scoped object

nabuskey:
We should handle the error returned by `CreateOrUpdate` to ensure the CLI does not hang indefinitely when error is returned. 

fixes: #94


GenPage (approved):
LGTM

jessesanford (approved):
LGTM

Issue #80, opened 2023-11-13
Deprecate and remove gitserver


nabuskey:
Once Gitea is implemented and tested, we should either mark gitserver deprecated and remove it. 

https://github.com/cnoe-io/idpbuilder/tree/main/pkg/controllers/gitserver

PR #97, opened 2023-11-28
prevent git server from running

nabuskey:
Currently the registry container is created and runs even though it is not used by anything. This PR prevents the git server controller from running and creating the registry container.

Part of: https://github.com/cnoe-io/idpbuilder/issues/80


jessesanford (approved):
LGTM

PR #101, opened 2023-12-05
Add delete and list command

greghaynes:
This isnt perfect, but this should handle the 90% case for when users dont want to have to install kind separately.

### DIFF
--- a/pkg/cmd/delete/root.go
+++ b/pkg/cmd/delete/root.go
@@ -1,0 +1,44 @@
+package delete
+
+import (
+	"flag"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	"sigs.k8s.io/kind/pkg/cluster"
+)
+
+var (
+	// Flags
+	buildName string
+)
+
+var DeleteCmd = &cobra.Command{
+	Use:   "delete",
+	Short: "Delete an IDP cluster",
+	Long:  ``,
+	RunE:  delete,
+}
+
+func init() {
+	DeleteCmd.PersistentFlags().StringVar(&buildName, "build-name", "localdev", "Name of the kind cluster to be deleted.")
+
+	zapfs := flag.NewFlagSet("zap", flag.ExitOnError)
+	opts := zap.Options{
+		Development: true,
+	}
+	opts.BindFlags(zapfs)
+	DeleteCmd.Flags().AddGoFlagSet(zapfs)
+
+	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+}
+
+func delete(cmd *cobra.Command, args []string) error {
+	provider := cluster.NewProvider(cluster.ProviderWithDocker())
+	if err := provider.Delete(buildName, ""); err != nil {
+		return errors.Wrapf(err, "failed to delete cluster %q", buildName)
+	}
+	return nil
+}
--- a/pkg/cmd/delete/root.go
+++ b/pkg/cmd/delete/root.go
@@ -3,9 +3,9 @@ package delete
 import (
 	"flag"
 
+	"github.com/cnoe-io/idpbuilder/pkg/cmd/helpers"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
-	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/kind/pkg/cluster"
 )
@@ -16,10 +16,11 @@ var (
 )
 
 var DeleteCmd = &cobra.Command{
-	Use:   "delete",
-	Short: "Delete an IDP cluster",
-	Long:  ``,
-	RunE:  delete,
+	Use:     "delete",
+	Short:   "Delete an IDP cluster",
+	Long:    ``,
+	RunE:    delete,
+	PreRunE: preDeleteE,
 }
 
 func init() {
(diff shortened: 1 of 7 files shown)


pkg/cmd/delete/root.go
```diff
@@ -0,0 +1,53 @@
...
+
+var DeleteCmd = &cobra.Command{
+	Use:   "delete",
+	Short: "Delete an IDP cluster",
+	Long:  ``,
+	RunE:  delete,
+}
+
+func init() {
+	DeleteCmd.PersistentFlags().StringVar(&buildName, "buildName", "localdev", "Name for build (Prefix for kind cluster name, pod names, etc).")
```
cmoulliard, line 27: The text "Name for build (Prefix for kind cluster name, pod names, etc)." is confusing. 
I suggest "Name of the kind cluster to be deleted."

pkg/cmd/list/root.go
```diff
@@ -0,0 +1,48 @@
...
+
+var ListCmd = &cobra.Command{
+	Use:   "list",
+	Short: "List idp clusters",
+	Long:  ``,
+	RunE:  delete,
+}
+
+func init() {
+	ListCmd.PersistentFlags().StringVar(&buildName, "buildName", "localdev", "Name of the kind cluster to be deleted.")
```
nabuskey, line 27: Should be `build-name` instead of `buildName` because we use kebab case everywhere else.

nabuskey (changes requested):

pkg/cmd/list/root.go
```diff
@@ -0,0 +1,48 @@
...
+var (
+	// Flags
+	buildName string
+)
+
+var ListCmd = &cobra.Command{
+	Use:   "list",
+	Short: "List idp clusters",
+	Long:  ``,
+	RunE:  delete,
```
nimakaviani, line 23: function name needs to be changed

pkg/cmd/list/root.go
```diff
@@ -0,0 +1,48 @@
...
+	opts := zap.Options{
+		Development: true,
+	}
+	opts.BindFlags(zapfs)
+	ListCmd.Flags().AddGoFlagSet(zapfs)
+
+	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+}
+
+func delete(cmd *cobra.Command, args []string) error {
```
nimakaviani, line 39: same as above

pkg/cmd/list/root.go
```diff
@@ -0,0 +1,48 @@
...
+
+var ListCmd = &cobra.Command{
+	Use:   "list",
+	Short: "List idp clusters",
+	Long:  ``,
+	RunE:  delete,
+}
+
+func init() {
+	ListCmd.PersistentFlags().StringVar(&buildName, "build-name", "localdev", "Name of the kind cluster to be deleted.")
```
nimakaviani, line 27: why require a `build-name` if listingg clusters? also the description needs fixing

nimakaviani (changes requested):

pkg/cmd/list/root.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func delete(cmd *cobra.Command, args []string) error {
+	provider := cluster.NewProvider(cluster.ProviderWithDocker())
+	clusters, err := provider.List()
+	if err != nil {
+		return errors.Wrapf(err, "failed to delete cluster %q", buildName)
+	}
+
+	fmt.Printf("Clusters: %v\n", clusters)
```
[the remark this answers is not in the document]
greghaynes: Whats that?

greghaynes:
@nimakaviani @nabuskey Sorry I dropped this - it should be ready for re-review

pkg/cmd/delete/root.go
```diff
@@ -0,0 +1,44 @@
...
+func init() {
+	DeleteCmd.PersistentFlags().StringVar(&buildName, "build-name", "localdev", "Name of the kind cluster to be deleted.")
+
+	zapfs := flag.NewFlagSet("zap", flag.ExitOnError)
+	opts := zap.Options{
+		Development: true,
+	}
+	opts.BindFlags(zapfs)
+	DeleteCmd.Flags().AddGoFlagSet(zapfs)
+
```
nabuskey, line 34: Let's use https://github.com/cnoe-io/idpbuilder/blob/8094f56b16d2babb945dd6caa0b67790fe928478/pkg/cmd/helpers/logger.go#L19 instead?

pkg/cmd/delete/root.go
```diff
@@ -0,0 +1,44 @@
...
+	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+}
+
+func delete(cmd *cobra.Command, args []string) error {
+	provider := cluster.NewProvider(cluster.ProviderWithDocker())
```
nabuskey, line 39: Wonder if we can use DetectNodeProvider() here? Not sure 100% but might be possible. Just in case someone created one with another provider. https://github.com/cnoe-io/idpbuilder/pull/182

cmoulliard:
Can you resolve the conflicts (rebase) please ?

cmoulliard (commented):
- Please rebase the code due to conflicts 
- Use a `printOutPut` function to list the clusters. Ideally the cluster listed should mention in separate columns: `core packages` and `custom` packages if available
```
// Example from list secrets
func printOutput(templatePath string, outWriter io.Writer, data []any, format string) error {
	switch format {
	case "json":
		b, err := json.MarshalIndent(data, "", "  ")
		if err != nil {
			return err
		}
		b = append(b, []byte("\n")...)
		_, err = outWriter.Write(b)
		return err
	case "yaml":
		b, err := yaml.Marshal(data)
		if err != nil {
			return err
		}
		_, err = outWriter.Write(b)
		return err
	case "":
		return renderTemplate(templatePath, outWriter, data)
	default:
		return fmt.Errorf("output format %s is not supported", format)
	}
}
```

greghaynes:
@cmoulliard I moved this to match Kind's upstream behavior - for a 'get clusters' command I'm not sure its necessary to have templates/json/etc. IME its pretty common to simply list the clusters (this is very useful for e.g. shell tooling integration). If we want to add extra formatting dont let me stop anyone from making that an optional argument.

greghaynes:
@nabuskey @nimakaviani Can you all re-review this please? thanks

pkg/cmd/get/clusters/clusters.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	"sigs.k8s.io/kind/pkg/cluster"
+)
+
+var ClustersCmd = &cobra.Command{
+	Use:   "clusters",
```
nimakaviani, line 15: not to be too picky here but looks like the other commands are verbs and this one is a noun. shall we keep it as `list` like it used to be? or `get` maybe?

nimakaviani (commented):
minor comment on the naming of the command, otherwise looks good.

nimakaviani:
@greghaynes once you resolve the conflict, this should be good to go.

nimakaviani (approved):
LGTM

nabuskey (approved):

PR #102, opened 2023-12-06
Remove enum validation for embedded app name

greghaynes:
Without this, not specifying embeeded app name errors because it does not adhere to the enum validation. I validated that changing the order does not work as well.

### DIFF
--- a/api/v1alpha1/gitrepository_types.go
+++ b/api/v1alpha1/gitrepository_types.go
@@ -16,8 +16,6 @@ type GitRepositorySpec struct {
 }
 
 type GitRepositorySource struct {
-	// +kubebuilder:validation:Optional
-	// +kubebuilder:validation:Enum:=argocd;backstage;crossplane;gitea;nginx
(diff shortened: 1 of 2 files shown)


nabuskey (commented):
We can remove enum but I thought it's a nice way to tell what apps are embedded instead of digging through the controller logic here: https://github.com/cnoe-io/idpbuilder/blob/9b25a0c62f25076d516b6a1b4217e930ae43bce6/pkg/controllers/localbuild/controller.go#L388

I think I should have specified the `omitempty` json tag when I implemented this. This was corrected in the custom package PR here: https://github.com/cnoe-io/idpbuilder/pull/100/files. It does work from my testing.

greghaynes:
oh interesting, I'll try that out - I thought omitempty was only for reading though (dont display an empty field)? I'm totally fine with enum as long as we can make it work, as is I am unable to set EmbeddedAppName to "" though, so as long as we enable this I'm happy.

nabuskey:
Closing this since it should be resolved in the PR: #100

PR #105, opened 2023-12-12
Update to GoLang 1.21

blakeromano:
This gets us using GoLang 1.21 which is the latest stable version of GoLang.


nimakaviani (approved):
LGTM

Issue #92, opened 2023-11-21
Print instructions on how to access ArgoCD UI


nabuskey:
We can improve user experience of using ArgoCD by making it easier for users to get the default admin credentials. 

We can either print the ArgoCD admin password to stdout or print the instructions on how to get the password e.g. `kubectl -n argocd get secret argocd-initial-admin-secret -o jsonpath="{.data.password}" | base64 -d` 

Whenever I spin up a new cluster, I have to search the command history to get the password and it gets a bit too annoying for me.

nimakaviani:
good idea!

cmoulliard:
Ideally we should get such information using the `idp` client and new commands:
```
./idpbuilder package list
./idpbuilder package show argocd
url: https://argocd.127.0.0.1.nip.io
user: admin
pwd: <password>
etc
./idpbuilder package show gitea
url: https://gitea.127.0.0.1.nip.io
user: admin
pwd: <password>
etc
```

PR #106, opened 2023-12-12
Print Argo Information After Creation

blakeromano:
Relates to https://github.com/cnoe-io/idpbuilder/issues/92

Makes it emit a series of log messages after the `idpbuilder create` command completes successfully. Gets the ArgoCD Password from K8s and logs the username/password and URL for ArgoCD.

Emits something like 

```
1.7024070614792352e+09  INFO    controller.custompackage        All workers finished    {"reconciler group": "idpbuilder.cnoe.io", "reconciler kind": "CustomPackage"}
1.702407061479521e+09   INFO    Stopping and waiting for caches
1.702407061479584e+09   INFO    Stopping and waiting for webhooks
1.7024070614795892e+09  INFO    Wait completed, proceeding to shutdown the manager


########################### Finished Creating IDP Successfully! ############################


Can Access ArgoCD at https://argocd.cnoe.localtest.me:8443/
Username: admin
Password can be retrieved by running: kubectl -n argocd get secret argocd-initial-admin-secret -o jsonpath="{.data.password}" | base64 -d
```

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -3,12 +3,15 @@ package build
 import (
 	"context"
 	"fmt"
+
 	"github.com/cnoe-io/idpbuilder/api/v1alpha1"
 	"github.com/cnoe-io/idpbuilder/globals"
 	"github.com/cnoe-io/idpbuilder/pkg/controllers"
 	"github.com/cnoe-io/idpbuilder/pkg/kind"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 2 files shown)


nimakaviani:
I think rather than printing the password for ArgoCD in plain text, it is a better practice to print the instructions on how to extract the password for ArgoCD. This also ensures that auth boundaries are respected for users, particularly when using remote repos

```bash
kubectl -n argocd get secret argocd-initial-admin-secret -o jsonpath="{.data.password}" | base64 -d
```

nimakaviani (changes requested):
see the comments above

blakeromano:
I hadn't thought of the remote cluster concept when I made this PR so that change totally makes sense! I updated the code accordingly and updated the comment with the new output

cmoulliard:
> it is a better practice to print the instructions on how to extract the password for ArgoCD.

I think that the idpbuilder CLI should propose a command that the users could use to get from the different core components the credentials and their URL to access them ;-)
```
idpbuilder config argocd
argocd url:
username:
password:

idpbuilder config gitea
gitea url
username:
password:
...
```

nimakaviani:
@nabuskey any particular reason why #92 only suggests to print ArgoCD info?

probably a better idea to share details on accessing all endpoints, for the embedded applciations (ArgoCD, Backstage, Gitea).

That said, I think we should print them all out after the run, but also we should have a separate command for `idpbuilder config` and `idpbuilder config -o json` that print config details in full.
 
with `idpbuilder config -o json` we will be able to feed the output to something like `jq` and make it usable for other scripts too.

nabuskey:
> @nabuskey any particular reason why #92 only suggests to print ArgoCD info?

I got frustrated getting ArgoCD creds repeatedly while debugging issues when I was implementing packaging. 

JSON output sounds good to me. What information do we want to print? While printing instructions for getting secret values itself is fine, it's probably useless for json output. Because if you return a command to run, you are asking the users to run arbitrary commands returned by another command.

nimakaviani:
idpbuilder will act based on the credentials available to it. probably fine when running `idpbuilder config -o json` or `idpbuilder config` to share base64 encoded creds directly as the output.

what we print though, at the end of the run, should still only be the instructions on how to get creds. This avoids data leaks as part of having the output printed in the CI etc.

nimakaviani:
@blakeromano given that your PR satisfies #92, we can merge this PR and create a separate issue to track the conversation in this thread. Alternatively, if you have the bandwidth to pick up the changes discussed here and revise your PR, that also would work. Let me know how you would want to proceed.

blakeromano:
I think for now let's go with merging this PR as it is (assuming we are happy with how it meets the criteria of #92) and create a separate issue.

nabuskey (approved):
LGTM and I agree this PR and what we discussed above should be separate.

nimakaviani (approved):
LGTM

PR #117, opened 2023-12-14
remove git server

nabuskey:
Git server is no longer used. 

closes: #80, #119


nimakaviani (changes requested):
Thank you, love seeing this happen.

I think we also need to clean up `pkg/apps` directory, since much of what is in there including the `Dockerfile`, the `entrypoint.sh` file and the configs for nginx in the Dockerfile etc are also there for the gitserver and need to go.

Also, what's under `api/v1alpha1` for gitserver needs to be removed.

`GitConfigSpec` stuff also needs to remove the optionality to choose between Gitea and GitServer https://github.com/cnoe-io/idpbuilder/blob/243ba5fd406cffa26b546cbb511d68ad1994dbd1/api/v1alpha1/localbuild_types.go#L23-L27

nabuskey:
I think I got them all. Please review when you get a chance.

nimakaviani (approved):
LGTM

PR #121, opened 2023-12-16
fix: use https for gitea

### LINKED ISSUE #113: Use https both internally and externally when using Gitea
Right now, the idpbuilder uses the local http connection on port 8880 when interacting with Gitea. This is where both http and https are available for Gitea and we use https over 8443 for all other services (backstage, argo, etc).

We should make it consistent with the rest of the services, and access everything over https. It is fine to use `skipVerify` when configuring the local go client for git in the idpbuilder code to connect to gitea.

nimakaviani:
- enable skipTLSVerify
- fix documentation

closes #113

### DIFF
--- a/pkg/controllers/gitrepository/controller.go
+++ b/pkg/controllers/gitrepository/controller.go
@@ -2,7 +2,9 @@ package gitrepository
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
+	"net/http"
 	"os"
 	"path/filepath"
 	"time"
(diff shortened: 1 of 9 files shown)


pkg/controllers/gitrepository/controller.go
```diff
@@ -13,7 +15,7 @@ import (
 	"github.com/cnoe-io/idpbuilder/pkg/util"
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/plumbing/object"
-	"github.com/go-git/go-git/v5/plumbing/transport/http"
+	gohttp "github.com/go-git/go-git/v5/plumbing/transport/http"
```
nabuskey, line 18: Why name it `gohttp`? Should be something like `gogithttp` if you want to be explicit.

nabuskey (approved):

pkg/controllers/gitrepository/controller.go
nimakaviani: fixed.

PR #124, opened 2023-12-26
ensure multi source spec is handled

nabuskey:
Fixes an issue when you are using multi sources in ArgoCD, the repo url was not updated correctly. It was because we were not updating the correct object.

### DIFF
--- a/pkg/controllers/custompackage/controller.go
+++ b/pkg/controllers/custompackage/controller.go
@@ -86,7 +86,7 @@ func (r *Reconciler) reconcileCustomPackage(ctx context.Context, resource *v1alp
 		synced := true
 		if app.Spec.HasMultipleSources() {
 			for j := range app.Spec.Sources {
-				s := app.Spec.Sources[j]
+				s := &app.Spec.Sources[j]
 				res, repo, sErr := r.reconcileArgocdSource(ctx, resource, appName, resource.Spec.ArgoCD.ApplicationFile, s.RepoURL)
 				if sErr != nil {
 					return res, sErr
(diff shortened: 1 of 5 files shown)


pkg/controllers/custompackage/test/resources/customPackages/testDir/app2.yaml
```diff
@@ -0,0 +1,19 @@
...
+  name: my-app2
+  namespace: argocd
+spec:
+  destination:
+    namespace: my-app2
+    server: "https://kubernetes.default.svc"
+  sources:
+    - repoURL: cnoe://busybox
+      targetRevision: HEAD
+      path: "."
```
cmoulliard, line 13: Sorry if I missed something but this test case only includes one source and not (as title mentions) several sources !

nabuskey: Yeah it's meant to represent when the sources field is used. Which is not exactly the same as what is described in the title. I'll update the test.

nabuskey:
Updated it to further handle a case where the same directory is specified twice in the spec. e.g.

```yaml
spec:
  sources:
    - repoURL: cnoe://app2
      targetRevision: HEAD
      path: "one"
    - repoURL: cnoe://app2
      targetRevision: HEAD
      path: "two"
```

cmoulliard (approved):

pkg/controllers/custompackage/controller.go
```diff
@@ -190,8 +190,13 @@ func (r *Reconciler) reconcileGitRepo(ctx context.Context, resource *v1alpha1.Cu
 		}
 		return nil
 	})
+	// it's possible for an application to specify the same directory multiple times in the spec.
+	// if there is a repository already created for this package, no further action is necessary.
+	if !errors.IsAlreadyExists(err) {
```
jessesanford, line 195: Good catch

jessesanford (approved):
LGTM

PR #128, opened 2024-01-16
ensure controllers have a chance to run at least once

nabuskey:
Currently, it is possible for the CLI to exit before controllers have a chance to act on changes. For example:

1. Run idp builder with custom package `./idpbuilder create --package-dir pkg/controllers/custompackage/test/resources/customPackages/testDir`. Then wait for it to finish.
2. Make modifications to manifests under `testDir`.
3. Run the same command again. Expectation is for everything to sync. Instead, you may get error messages like: 
    ```
    1.7054368071870215e+09	ERROR	controller.gitrepository	failed updating repo status	{"reconciler group": "idpbuilder.cnoe.io", "reconciler kind": "GitRepository", "name": "my-app2-app2", "namespace": "idpbuilder-localdev", "error": "client rate limiter Wait returned an error: context canceled"}
    github.com/cnoe-io/idpbuilder/pkg/controllers/gitrepository.(*RepositoryReconciler).Reconcile
	    /home/ubuntu/idpbuilder/pkg/controllers/gitrepository/controller.go:114
    sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile
	    /home/ubuntu/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.0/pkg/internal/controller/controller.go:114
    sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	    /home/ubuntu/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.0/pkg/internal/controller/controller.go:311
    sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	    /home/ubuntu/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.0/pkg/internal/controller/controller.go:266
    sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
	    /home/ubuntu/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.0/pkg/internal/controller/controller.go:227
    ```

This is because we do not check if custom package and git repo controllers have worked on objects before exiting CLI.

We should use the commonly used `observedGeneration` field in Status, then check if the value matches with the manifest generation value before exiting.

For manifest generation to update, the spec field must be updated. I have introduced a new optional field for each type that gets populated with a random value every CLI run to force generation update.

Instead of expanding specs, we could annotate the objects with run ID. Then, controllers can annotate them with a known value once they are done working on them. This approach seemed less transparent so I went with generation instead.


nimakaviani (commented):
Thanks, the PR generally looks good to me.

I am however not convinced on the transparency argument. Why does the transparency in having the runId as part of the spec matter to the developers? sure I can do a `kubectl get` and see it in the spec but I can see it as an annotation too.

As an end user of the tool it is more confusing if I check the Spec for the resource and I am not sure what the value for the field in the spec is or what purpose it has. Also I think what we put in the spec becomes part of the API, where an input is expected. Which is why I am more hesitant here.

nabuskey:
I don't really care which way we go. It's just that observedGeneration was perfect for this kind of uses so I wanted to stay closer to general practices, instead of inventing our own annotations.  I'll re-work it.

jessesanford:
> As an end user of the tool it is more confusing if I check the Spec for the resource and I am not sure what the value for the field in the spec is or what purpose it has. Also I think what we put in the spec becomes part of the API, where an input is expected. Which is why I am more hesitant here.
> 

This is exactly what I would think. Why not make it an annotation and maybe instead of making it a random value make it the exact execution time of the cli run with enough precision to ensure uniqueness? Maybe we can call it cli-creation-time or something?

jessesanford (commented):
Looks good but see comments

nabuskey:
I agree that it's a better user experience, and uniqueness guarantee is probably not needed. Re-worked it to use annotations with timestamp (RFC3339Nano) instead. 

Introduces two annotations. 
- `cnoe.io/cli-start-time` Contains timestamp that gets updated by CLI on startup.
- `cnoe.io/last-observed-cli-start-time` Gitrepo and custom pkg controllers updates this value to match `cnoe.io/cli-start-time`value.

nimakaviani (approved):
Nice, LGTM

PR #131, opened 2024-01-23
update default cluster version to 1.27

nabuskey:
We use K8s version 1.27 in tests, while the default version in CLI is set at 1.26. We should use 1.27 for all.

1.27.3 was taking from Kind release that we use: https://github.com/kubernetes-sigs/kind/releases/tag/v0.20.0

### DIFF
--- a/pkg/cmd/create/root.go
+++ b/pkg/cmd/create/root.go
@@ -37,7 +37,7 @@ var CreateCmd = &cobra.Command{
 func init() {
 	CreateCmd.PersistentFlags().BoolVar(&recreateCluster, "recreate", false, "Delete cluster first if it already exists.")
 	CreateCmd.PersistentFlags().StringVar(&buildName, "build-name", "localdev", "Name for build (Prefix for kind cluster name, pod names, etc).")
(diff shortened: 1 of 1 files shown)


blakeromano (commented):
I'd wonder if we should just go to 1.28?

nabuskey:
We could. But our controller-runtime is a bit old and technically they are not compatible as-is. 1.27 seems ok from my testing. I think I am more inclined to upgrade to 1.28 once we have the controller-runtime dependencies figured out.

blakeromano:
> We could. But our controller-runtime is a bit old and technically they are not compatible as-is. 1.27 seems ok from my testing. I think I am more inclined to upgrade to 1.28 once we have the controller-runtime dependencies figured out.

I think as long as that is apart of the controller-runtime dependency work then that sounds fine!

blakeromano (approved):

cmoulliard:
Is it documented on the README.md that idpbuilder should be deployed on kind >= 1.27 since release 0.x ?
@nabuskey

greghaynes:
Lets just merge this - I dont think theres a user facing version dependency here we need to worry about documenting since we vendor our kind dep. If a user has a different kind version IME it "just works".

greghaynes (approved):

PR #141, opened 2024-01-31
Fix #84: remove GitConfig from API spec

### LINKED ISSUE #84: Use KubeBulder validations to verify git type
Ensure that in the LocalBuild config, `GitConfig.Type` can be set only from an enum list of `gitserver` and `gitea`

https://github.com/cnoe-io/idpbuilder/blob/11a9914536cbd8ee8da9f5d79e8924f1e595ec0c/api/v1alpha1/localbuild_types.go#L26

hodgiwabi:


nimakaviani:
This could be a left over from when we did #117 

We removed the gitserver all together. so the enum in the spec is not relevant anymore. 

@hodgiwabi, do you mind updating the PR to remove that field from the spec? thank you!

hodgiwabi:
@nimakaviani makes sense to me! updated to remove gitConfig from the spec.

nabuskey (approved):
LGTM!

Issue #95, opened 2023-11-22
Upgrade controller-runtime


nabuskey:
We are using an older version of controller-runtime. 

We should upgrade to one of latest ones to avoid running into old bugs and keep dependencies somewhat close.

nabuskey:
Looks like we cannot upgrade because we have to import ArgoCD.

```
# k8s.io/kubernetes/pkg/apis/core/v1
../go/pkg/mod/k8s.io/kubernetes@v1.24.2/pkg/apis/core/v1/zz_generated.conversion.go:5192:22: cannot use (*v1.TypedLocalObjectReference)(unsafe.Pointer(in.DataSourceRef)) (value of type *"k8s.io/api/core/v1".TypedLocalObjectReference) as *"k8s.io/api/core/v1".TypedObjectReference value in assignment
```

We either need to contribute to ArgoCD and make types its own package or replicate types in our repo.

PR #145, opened 2024-02-07
remove direct argocd package dependency

nabuskey:
This PR removes direct dependency to ArgoCD. This is necessary because we need to import ALL ArgoCD to use ArgoCD types. Importing all ArgoCD causes dependency issues as described here: 

- https://github.com/cnoe-io/idpbuilder/issues/95#issuecomment-1858214772
- https://cloud-native.slack.com/archives/C05TN9WFN5S/p1706137401642309


jessesanford (commented):
LGTM

greghaynes (approved):

PR #146, opened 2024-02-09
Fix #108: Add version command

### LINKED ISSUE #108: add a Version command
Right now, when tracking issues with idpBuilder, it is hard to reason about what version of the tool is used by the users.

It would be helpful for an `idpbuilder version` command to be added that corresponds to the released version of the idpbuilder.

Also, running `make build` in the terminal, should add a suffix to the version, indicating a local build, while `make build` in the CI and as part of the release cycle, should bump the version and remove any suffix from it, indicating a clean ready to release build of the tool

hodgiwabi:
Adds a simple version command using ldflags on build and `runtime` package at runtime.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -13,7 +13,7 @@ builds:
       - amd64
       - arm64
     ldflags:
-      - -X github.com/cnoe-io/idpbuilder/pkg/cmd/version.idpbuilderVersion=$(shell git describe --tags --dirty --broken)
+      - -X github.com/cnoe-io/idpbuilder/pkg/cmd/version.idpbuilderVersion=$(shell git describe --always --tags --dirty --broken)
       - -X github.com/cnoe-io/idpbuilder/pkg/cmd/version.gitCommit=$(shell git rev-parse HEAD)
       - -X github.com/cnoe-io/idpbuilder/pkg/cmd/version.buildDate=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
     binary: idpbuilder
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -13,9 +13,9 @@ builds:
       - amd64
       - arm64
     ldflags:
-      - -X github.com/cnoe-io/idpbuilder/pkg/cmd/version.idpbuilderVersion=$(shell git describe --always --tags --dirty --broken)
-      - -X github.com/cnoe-io/idpbuilder/pkg/cmd/version.gitCommit=$(shell git rev-parse HEAD)
-      - -X github.com/cnoe-io/idpbuilder/pkg/cmd/version.buildDate=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
+      - -X github.com/cnoe-io/idpbuilder/pkg/cmd/version.idpbuilderVersion={{ .Version }}
+      - -X github.com/cnoe-io/idpbuilder/pkg/cmd/version.gitCommit={{ .FullCommit }}
+      - -X github.com/cnoe-io/idpbuilder/pkg/cmd/version.buildDate={{ .CommitDate }}
     binary: idpbuilder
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -10,14 +10,15 @@ require (
 	github.com/docker/docker v24.0.6+incompatible
 	github.com/go-git/go-git/v5 v5.10.0
 	github.com/google/go-cmp v0.6.0
-	github.com/spf13/cobra v1.7.0
+	github.com/spf13/cobra v1.8.0
```
nabuskey, line 13: Is this necessary?

nabuskey (commented):
For the version command, the most relevant time and place is when executed at end users' local machines. So we need to make sure version information is propagated to released binaries. 

There's a bit of complication here. We use go releaser to do the bulk of the work to cut releases, and It supports importing pre-built binaries to the workflow. e.g. Use make to build, then use built binaries for the rest of the steps of cutting a new release. This is a perfect use case for us. 
Unfortunately, this is a paid feature and we currently do not pay for it. 

This means we need to put these information in the go releaser configuration as well. https://github.com/cnoe-io/idpbuilder/blob/main/.goreleaser.yaml

It looks like it does have support for it ld flags. https://goreleaser.com/customization/builds/

go.mod
hodgiwabi: No I don't believe it is. I think I slipped the `go get -u` in on accident--I'll undo it if that's preferred.

hodgiwabi:
No worries! I'll update the goreleaser as well and see if I can test that out locally as well.

pkg/cmd/version/root.go
```diff
@@ -0,0 +1,112 @@
...
+
+func init() {
+	VersionCmd.Flags().StringVarP(&outputFormat, "output", "o", "", `Print the idpbuilder version information in a given output format. Accepts "wide", "json", and "yaml".`)
+}
+
+var (
+	idpbuilderVersion = "unknown"
+	goVersion         = runtime.Version()
+	goOs              = runtime.GOOS
+	goArch            = runtime.GOARCH
```
hodgiwabi, line 32: *Note:*
Using the `runtime` package here [like Kind does](https://github.com/kubernetes-sigs/kind/blob/f7e5c8e407269a6d5b7a6f86e6b3f9c13e9d8e9b/pkg/cmd/kind/version/version.go#L53)

Makefile
```diff
@@ -1,10 +1,15 @@
+LD_FLAGS=-ldflags " \
+    -X github.com/cnoe-io/idpbuilder/pkg/cmd/version.idpbuilderVersion=$(shell git describe --always --tags --dirty --broken) \
```
hodgiwabi, line 2: `--always` flag means that this won't fail with forks/local dev and not output anything for `idpbuilderVersion` if there are no local tags.

.goreleaser.yaml
```diff
@@ -12,6 +12,10 @@ builds:
     goarch:
       - amd64
       - arm64
+    ldflags:
+      - -X github.com/cnoe-io/idpbuilder/pkg/cmd/version.idpbuilderVersion={{ .Version }}
+      - -X github.com/cnoe-io/idpbuilder/pkg/cmd/version.gitCommit={{ .FullCommit }}
+      - -X github.com/cnoe-io/idpbuilder/pkg/cmd/version.buildDate={{ .CommitDate }}
```
hodgiwabi, line 18: Using the [Name Templates](https://goreleaser.com/customization/templates/) provided in Goreleaser for the release flags, for consistency.

hodgiwabi:
@nabuskey Ready for review, barring the tests passing. Not sure if that's related to this PR or not.

locally, this works really well, and I believe this will work well in CI. I don't know if I can add tests for this command effectively... do you know where those might go if I could add them? or if there are any tests for the `create` command?

local test results:
image

PR #147, opened 2024-02-09
upgrade controller-runtime and sync k8s package versions

nabuskey:
This upgrades controller-runtime to v0.17.1. This allows us to use slog for logging as well. 

I've also updated k8s.io package versions to be the same. Otherwise it does not compile. I think these replace statements were needed because we had hard dependencies on ArgoCD. 

We really need to update our default k8s version. Version skew is becoming too big for my liking.

closes: #95


pkg/controllers/gitrepository/controller_test.go
```diff
@@ -7,10 +7,11 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
-	ctrl "sigs.k8s.io/controller-runtime"
 	"testing"
 	"time"
 
+	ctrl "sigs.k8s.io/controller-runtime"
+
```
nimakaviani, line 14: necessary?

go.mod
```diff
@@ -67,43 +67,42 @@ require (
...
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
-	github.com/prometheus/client_golang v1.16.0 // indirect
-	github.com/prometheus/client_model v0.3.0 // indirect
-	github.com/prometheus/common v0.42.0 // indirect
-	github.com/prometheus/procfs v0.10.1 // indirect
+	github.com/prometheus/client_golang v1.18.0 // indirect
+	github.com/prometheus/client_model v0.5.0 // indirect
+	github.com/prometheus/common v0.45.0 // indirect
+	github.com/prometheus/procfs v0.12.0 // indirect
```
nimakaviani, line 87: love these changes.

nimakaviani (approved):
LGTM

pkg/controllers/gitrepository/controller_test.go
nabuskey: Different go import sorting. We aren't being consistent at all right now but we should standardize on goimports

PR #148, opened 2024-02-15
(feat): Add codespaces/devpod support

csantanapr:
Adds support to be able to run idpbuilder on cloud dev environment


.devcontainer/postCreateCommand.sh
```diff
@@ -0,0 +1,36 @@
...
+sudo mv linux-${TARGETARCH}/helm /usr/local/bin && rm helm3.tar.gz && rm -R linux-${TARGETARCH}
+
+# Install kubebuilder
+curl -L -o kubebuilder "https://go.kubebuilder.io/dl/latest/${TARGETOS}/${TARGETARCH}" && chmod +x kubebuilder
+sudo mv kubebuilder /usr/local/bin/
+
+
+# install protocol buffer compiler (protoc)
+sudo apt update
+sudo apt install -y protobuf-compiler
```
nabuskey, line 26: We don't need these. We do need `build-essential` if not installed already.

nabuskey (commented):
Thank you for this! Left some comments, please take a look

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,26 @@
+{
+  "image": "mcr.microsoft.com/vscode/devcontainers/base:ubuntu",
+  "features": {
+    "ghcr.io/devcontainers/features/go:1": {
+      "version": "1.21"
+    },
+    "ghcr.io/devcontainers/features/node:1": {
+      "version": "20"
+    },
```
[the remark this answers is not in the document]
csantanapr: I think is useful as a full IDE to have to have `node` and `python` in addition of `go` as we use the same IDE devcontainer to test examples

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,26 @@
...
+    "ghcr.io/devcontainers/features/docker-in-docker:2": {},
+    "ghcr.io/devcontainers/features/python:1": {}
```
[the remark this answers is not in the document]
csantanapr: I think is useful as a full IDE to have to have `node` and `python` in addition of `go` as we use the same IDE devcontainer to test examples

.devcontainer/postCreateCommand.sh
csantanapr: @nabuskey how do I check if its already installed. I was able to do `make` and build the binary

.devcontainer/postCreateCommand.sh
```diff
@@ -0,0 +1,42 @@
...
+set -eux
+
+TARGETOS=linux
+TARGETARCH=amd64
+
+# Install Kind
+curl -L -o kind "https://kind.sigs.k8s.io/dl/v0.22.0/kind-linux-${TARGETARCH}" && chmod +x ./kind
+sudo mv ./kind /usr/local/bin
+
+# Install kubectl
```
sunnygoel87, line 11: Isn't `kubectl` installed by default on codespace? This is what I see when I forked `idpbuilder` repo and created a codespace .

```
@sunnygoel87 ➜ /workspaces/idpbuilder (main) $ kubectl version
Client Version: v1.29.1
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
The connection to the server localhost:8080 was refused - did you specify the right host or port?
```

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,26 @@
...
+  },
+  "forwardPorts": [
+    8443
+  ],
+  "hostRequirements": {
+    "cpus": 4
```
sunnygoel87, line 17: Do we really need 4 cpus to run idp in codespace? The reason I ask, GitHub Free personal account has 120 core compute hours quota per month so if we don't need 4 cpus then we can reduce the compute capacity and buy more time to play with idp within free quota range

.devcontainer/postCreateCommand.sh
```diff
@@ -0,0 +1,42 @@
...
+curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/${TARGETARCH}/kubectl && chmod +x ./kubectl
+sudo mv ./kubectl /usr/local/bin
+
+# Install helm
```
sunnygoel87, line 15: Do we need to install `helm` ? Because this is what I see when I ran `helm version` command on forked repo in the codespace.

```
sunnygoel87 ➜ /workspaces/idpbuilder (main) $ helm version
version.BuildInfo{Version:"v3.14.0", GitCommit:"3fc9f4b2638e76f26739cd77c7017139be81d0ea", GitTreeState:"clean", GoVersion:"go1.21.5"}
```

.devcontainer/postCreateCommand.sh
```diff
@@ -0,0 +1,42 @@
...
+bash -c "curl -s https://get.helm.sh/helm-v3.14.1-linux-${TARGETARCH}.tar.gz > helm3.tar.gz" && tar -zxvf helm3.tar.gz linux-${TARGETARCH}/helm && chmod +x linux-${TARGETARCH}/helm
+sudo mv linux-${TARGETARCH}/helm /usr/local/bin && rm helm3.tar.gz && rm -R linux-${TARGETARCH}
+
+# Install kubebuilder
```
sunnygoel87, line 19: I thought, this was a turnkey solution to setup idp on local machine. What exactly are we planning to build using `kubebuilder` ?

.devcontainer/postCreateCommand.sh
```diff
@@ -0,0 +1,42 @@
...
+curl -L -o kubebuilder "https://go.kubebuilder.io/dl/latest/${TARGETOS}/${TARGETARCH}" && chmod +x kubebuilder
+sudo mv kubebuilder /usr/local/bin/
+
+# Install Github CLI
```
sunnygoel87, line 23: Do we need to install github CLI ? Because this is what I see when I ran `gh version` command on forked repo in the codespace.

```
@sunnygoel87 ➜ /workspaces/idpbuilder (main) $ gh version
gh version 2.43.1 (2024-01-31)
https://github.com/cli/cli/releases/tag/v2.43.1
```

.devcontainer/devcontainer.json
csantanapr: your right I will remove the cpus requirement

.devcontainer/postCreateCommand.sh
csantanapr: The repo is to build idpbuilder, idpbuilder uses control-runtime. I don't think we need this I will remove

.devcontainer/postCreateCommand.sh
csantanapr: I'm using devpod with the base image of `"image": "mcr.microsoft.com/vscode/devcontainers/base:ubuntu",` specify in the devcontainers.json

csantanapr: Why I'm including because is useful when testing and reviewing PRs by copy and paste like `gh pr checkout 148`

.devcontainer/postCreateCommand.sh
csantanapr: @sunnygoel87 it depends what is the based image your using in codespaces.
I added a check to see if helm is not present and install it then. I'm not using codespaces I'm using [devpod](https://devpod.sh/)

csantanapr:
I had to remove forwardPorts from devcontainers.json as this prevented exposing port 8843 in localhost, it wanted only to expose to <unique>.github.dev host name. forwardPorts only worked for devpod. 
User in both cases need to go to "ports" in the local vscode/ide and go to Ports and add forward port 8443 to 127.0.0.1
User will also need to add the /etc/hosts entry locally in laptop/workstation
```
127.0.0.1 argocd.cnoe.localtest.me
127.0.0.1 backstage.cnoe.localtest.me
127.0.0.1 gitea.cnoe.localtest.me
127.0.0.1 argo.cnoe.localtest.me
127.0.0.1 keycloak.cnoe.localtest.me
```
and any other hostname handle by nginx ingress

.devcontainer/postCreateCommand.sh
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env sh
+
+TARGETARCH=amd64
+
+# Install Kind
+if ! command -v kubectl &> /dev/null; then
```
sunnygoel87, line 6: @csantanapr - should it be `command -v kind` instead of `command -v kubectl` at this point?

nabuskey:
Some things I've noticed about trying to get it working in codespaces: 

1. Codespaces creates a public proxy endpoint for processes running in codespaces. This is considered proxy dodging and very likely gets blocked by security software managed by organizations on local machines. 
2. Because of above, ingress-nginx cannot route based on host names. For example, when I port-forward 8443, it creates an endpoint looking something like `https://random-name-8443.app.github.dev`.  To access the port 8443 on codespace, we must use this url.  This means we must use path based routing in our ingress instead of hostname based routing. e.g. instead of using `argocd.localtest.me`, do `random-name-8443.app.github.dev/argocd`. 
3. This means we need to configure argocd to append `/argocd` to its root path. Otherwise, the client can't fetch resources from ArgoCD. 

So if we want to get this to work in Codespaces flawlessly, we need to configure our core packages for this purpose specifically.

nabuskey:
> User will also need to add the /etc/hosts entry locally in laptop/workstation
>
> ```
>127.0.0.1 argocd.cnoe.localtest.me
>127.0.0.1 backstage.cnoe.localtest.me
>127.0.0.1 gitea.cnoe.localtest.me
>127.0.0.1 argo.cnoe.localtest.me
>127.0.0.1 keycloak.cnoe.localtest.me
>```

This shouldn't be necessary and we probably should not ask users to do this because organization security policies may not allow this kind of modifications. The domain names `*.localtest.me` does resolve to 127.0.0.1 as long as the registered NS is reachable.

csantanapr:
@nabuskey I agree it would be better experience to have idpbuilder work with the external hostname given by Codespaces.
Reading the docs is not a random-name but the `https://CODESPACENAME-PORT.app.github.dev` according to the docs [1] the `CODESPACENAME` is a environment `CODESPACE_NAME` available from the environment [2]

I think is matter telling the user to find and replace in the yamls `host: argo.cnoe.localtest.me` with `host: $CODESPACE_NAME` 

I will test this setup

[1]: https://docs.github.com/en/codespaces/developing-in-a-codespace/forwarding-ports-in-your-codespace
[2]: https://docs.github.com/en/codespaces/developing-in-a-codespace/default-environment-variables-for-your-codespace#list-of-default-environment-variables

csantanapr:
I think this is ready for merging, is a very basic dev environment for go, and if user wants to use idpbuilder is also has docker in docker that user can use idpbuilder

nabuskey (changes requested):

.devcontainer/postCreateCommand.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env sh
+
+curl -sLS https://get.arkade.dev | sudo sh
```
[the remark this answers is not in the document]
csantanapr: arkade is useful if you need to install later more clis
I had a shell and curl, but I had to account for architecture
Can change it to curl and assume amd64

.devcontainer/postCreateCommand.sh
```diff
@@ -0,0 +1,21 @@
...
+arkade get kind
+arkade get kubectl
+arkade get helm
+echo "export PATH=\$PATH:/home/vscode/.arkade/bin" >> $HOME/.bashrc
+
+# Make sure go path is owned by vscode
+sudo chown -R vscode:vscode /home/vscode/go
+
+# Add idpbuilder to path
+echo "export PATH=\$PATH:/home/vscode/go/src/github.com/cnoe-io/idpbuilder" >> $HOME/.bashrc
```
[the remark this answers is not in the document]
csantanapr: Is convenient if used wants to `make build` follow by `idpbuilder`
Don't think we want to download the latest release, this devcontainer is for code contributions
I can remove it if you think user running `./idpbuilder` is better

csantanapr:
@nabuskey please take a look I addressed your feedback. The files now contain the absolute minimum for an user to use codespaces, devpod, docker, to have a golang environment to make changes to idpbuilder and test

nimakaviani:
LGMT!

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,15 @@
...
+  "image": "mcr.microsoft.com/vscode/devcontainers/base:ubuntu",
+  "features": {
+    "ghcr.io/devcontainers/features/go:1": {
+      "version": "1.21"
+    },
+    "ghcr.io/devcontainers/features/docker-in-docker:2": {}
+  },
+  "postCreateCommand": ".devcontainer/postCreateCommand.sh",
+  "workspaceMount": "source=${localWorkspaceFolder},target=/home/vscode/go/src/github.com/cnoe-io/idpbuilder,type=bind",
+  "workspaceFolder": "/home/vscode/go/src/github.com/cnoe-io/idpbuilder",
```
nabuskey, line 11: Does this have to be under the default GOPATH? I would prefer to have this directly under the user's home path. e.g. `/home/vscode/workspaces/idpbuilder`

csantanapr: @nabuskey please take another look when you have a chance, I simplify the setup to the bare minimal now, I think

nabuskey (approved):
LGTM

PR #149, opened 2024-02-15
We no longer install backstage by default so remove this from README

greghaynes:


nabuskey (approved):
LGTM

cmoulliard:
If we don't install backstage by default, what is then the added value of the IDPBuilder tool ? If now idp builder creates a kind cluster to install argocdCD, ingress, gitea, then it should be renamed to `gitopsBuilder` !

@nabuskey @nimakaviani

nimakaviani:
@cmoulliard the point here is that we are not making backstage an embedded package. Our reference implementation however should continue and support inclusion of Backstage as the developer portal (but via a custom package installation)

There are several reasons to avoid embedding backstage, one key one being that, users often have their custom configuration and images that they want to bring in for Backstage, and tying them back to embedded configurations just makes it harder for them to customize the deployment.

I believe we had an extensive conversation around this either on slack or on another issue in this repo, but I wasnt able to find it. I will post it here for the records, if I manage to find it.

PR #153, opened 2024-02-20
feat: change from port 8443 to 9443

csantanapr:
This allows to use Codespaces for port forward thru ssh in VS Code Desktop. The port `8443` it doesn't allow to map to localhost/127.0.0.1


nabuskey:
Closing this because we can configure ports in CLI.

PR #155, opened 2024-02-24
make the serving port for idpBuilder configurable

### LINKED ISSUE #139: Configurable application port
Currently, we listen on port 8443 for TLS traffic because the common 443 port is considered privileged and may not work for all users. This does not mean this value should be immutable. I think we should allow end users to customize this. 

We could possibly provide a new flag `--port`. When specified, it must change a few modifications. 

1. https://github.com/cnoe-io/idpbuilder/blob/5194086e5d2f087682d13ceb659d4b8433cee006/pkg/controllers/localbuild/gitea.go#L18
2. https://github.com/cnoe-io/idpbuilder/blob/5194086e5d2f087682d13ceb659d4b8433cee006/pkg/kind/resources/kind.yaml#L16

nimakaviani:
- add a --port flag to configure the default port
- add go templates directives to override embedded resource manifests
- template kind cluster config
- add a sample script to tweak the port for the reference example

closes #139

### DIFF
--- a/examples/local-backup/kind.yaml
+++ b/examples/local-backup/kind.yaml
@@ -15,5 +15,5 @@ nodes:
       containerPath: /backup
   extraPortMappings:
   - containerPort: 443
-    hostPort: 8443
+    hostPort: {{ .Port }}
     protocol: TCP
--- a/examples/ref-implementation/README.md
+++ b/examples/ref-implementation/README.md
@@ -26,6 +26,8 @@ idpbuilder create --package-dir examples/ref-implementation
 
 This will take ~6 minutes for everything to come up. To track the progress, you can go to the [ArgoCD UI](https://argocd.cnoe.localtest.me:8443/applications).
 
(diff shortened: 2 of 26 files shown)


nimakaviani:
This PR only dynamically changes the port for the embedded tools and configurations. For external configurations (e.g. the reference implementation) it is left up to the users to make the required modifications to change the port. 

For the reference example implementation, the PR also includes a small script that can help users tweak the port prior to deploying the reference implementation. This should help in places such as codespaces where there is a desire to use a port different from the default 8443 port.

nabuskey (approved):
LGTM

cmoulliard:
I know that most of the URLs that we are proposing should be exposed using HTTPS/TLS (= `8443` or `443`) BUT ideally we should also allow to configure the port `80` like this 
```
Source: https://github.com/snowdrop/k8s-infra/blob/main/kind/kind.sh#L505-L512
  - containerPort: ${INGRESS_80_CONTAINER_PORT}
    hostPort: ${INGRESS_80_CONTAINER_PORT}
    protocol: TCP
    listenAddress: "0.0.0.0"
  - containerPort: ${INGRESS_443_CONTAINER_PORT}
    hostPort: ${INGRESS_443_CONTAINER_PORT}
    protocol: TCP
    listenAddress: "0.0.0.0"
```
@nimakaviani

cmoulliard:
> any particularly usecase @cmoulliard where we cant get away with 443 and require 80 to be open as well?

No. We need both ports and to be able to setup: 80 and 443, etc as ingress's proxy handle HTTP or HTTPS traffic ;-)

Issue #49, opened 2023-10-19
Use docker testcontainer able to launch: local docker registry, etc for testing cases


cmoulliard:
## Proposition

I would like to propose to use this project https://golang.testcontainers.org/quickstart/ for our integration tests when it is needed to launch some containers. Example: To execute this `image_test.go` test, a local registry container is needed

We can easily use testcontainer to bootstrap a container as you can see hereafter:

```
package gitserver

import (
	"context"
	"github.com/cnoe-io/idpbuilder/api/v1alpha1"
	"github.com/cnoe-io/idpbuilder/pkg/apps"
	"github.com/cnoe-io/idpbuilder/pkg/docker"
	"github.com/docker/docker/api/types"
	"github.com/docker/docker/api/types/container"
	"github.com/docker/go-connections/nat"
	"github.com/testcontainers/testcontainers-go"
	"github.com/testcontainers/testcontainers-go/wait"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	controllerruntime "sigs.k8s.io/controller-runtime"
	logf "sigs.k8s.io/controller-runtime/pkg/log"
	"sigs.k8s.io/controller-runtime/pkg/log/zap"
	"strings"
	"testing"
)

func TestReconcileGitServerImage(t *testing.T) {
	logf.SetLogger(zap.New(zap.UseDevMode(true)))

	ctx := context.Background()
	req := testcontainers.ContainerRequest{
		Image:      "registry:2",
		WaitingFor: wait.ForListeningPort("5000/tcp"),
		HostConfigModifier: func(hc *container.HostConfig) {
			hc.PortBindings = nat.PortMap{
				"5001/tcp": []nat.PortBinding{
					{
						HostIP:   "localhost",
						HostPort: "5001",
					},
				},
			}
		},
	}
	registryC, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
		ContainerRequest: req,
		Started:          true,
	})
	if err != nil {
		panic(err)
	}
	defer func() {
		if err := registryC.Terminate(ctx); err != nil {
			panic(err)
		}
	}()
```

there is even a trick to bind the needed port (e.g 5001) to the port used by the docker registry: `5000`

Test output will be now
```
[...]

PR #156, opened 2024-02-26
Add docker-registry example package

greghaynes:


nimakaviani:
very cool. to some extent addresses this issue as well: https://github.com/cnoe-io/idpbuilder/issues/49

nabuskey:
Can you add a readme similar to other examples we have?

greghaynes:
@jessesanford I agree with @nabuskey we should add a readme. One thing I would definitely include in there is this: https://engineering.autodesk.com/summits/techx2024/proposals/view/87157 - we set a default password and what it is.

csantanapr (changes requested):
Please a hello world in a `README.md` at the root of the new example folder

examples/ref-implementation/backstage/manifests/install.yaml
```diff
@@ -166,7 +166,7 @@ data:
       locations:
         # Examples from a public GitHub repository.
         - type: url
-          target: https://gitea.cnoe.localtest.me:8443/giteaAdmin/idpbuilder-localdev-backstage-entities/src/branch/main/catalog-info.yaml
+          target: https://gitea.cnoe.localtest.me:8443/giteaAdmin/idpbuilder-localdev-backstage-templates-entities/raw/branch/main/catalog-info.yaml
```
csantanapr, line 169: rebase this is already fix in main
https://github.com/cnoe-io/idpbuilder/commit/ea9c56e6231c05cc33ed6b33a00bec79008a8a16

jessesanford: Yes this is the cherry-pick of that exact commit to this branch I needed it to make sure things were working correctly. The merge to main should be noOp here. Happy to remove it if that is more appropriate.

omrishiv:
Do we want to use gitea's container registry or maybe [harbor](https://github.com/goharbor/harbor)? gitea's would mean not having to deploy an extra resource into the cluster whereas harbor provides supply chain security and image scanning, which could be a nice addition

cmoulliard:
> Do we want to use [gitea's container registry](https://docs.gitea.com/usage/packages/container) or maybe [harbor](https://github.com/goharbor/harbor)? gitea's would mean not having to deploy an extra resource into the cluster whereas harbor provides supply chain security and image scanning, which could be a nice addition

As the idpbuilder's gitea server can also be used as images registry, such a package is not needed OR could be proposed as separated ref implementation example

nimakaviani:
yeah closing this as #282 takes care of this. Feel free to reopen if you think a separate solution is needed.

PR #164, opened 2024-03-01
support path based routing in core packages

nabuskey:
This introduces new flags that can be used when creating a new cluster. 

- `--protocol`. Either http or https. We have been using https but http is useful when we want to front it with other proxy services like codespaces.
- `--host`. Specifies the host name that ingress nginx should use to route requests. When fronted by another proxy, you can set this to match the host name of the proxy, if required.
- `--use-path-routing`. Allows you to have core packages under a single domain name. e.g. instead of going to `https://argocd.cnoe.localtest.me`, go to `https://cnoe.localtest.me/argocd`. This flag applies to core packages only.


greghaynes:
@nabuskey interesting. I dont understand what on codespaces breaks this. Is there a doc or a hint you can leave about how it is different?

nabuskey:
@greghaynes See my comment here: https://github.com/cnoe-io/idpbuilder/pull/148#issuecomment-1952942333

It pretty much comes down to many web based IDE environments not supporting raw TCP connection forwarding. It makes sense that browsers don't support it,  but it makes running and interacting with idpbuilder in a browser impossible. I think having full support for Codespaces is quite useful for letting new users play with idpbuidler and realize the benefit of it quickly.

greghaynes:
Ah that makes sense! I wonder if we should just switch everything to be path based, then? I'm fine with this regardless, thanks!

greghaynes (approved):

nabuskey:
Yeah that's what I was discussing with Nima earlier. I wonder if it makes sense to switch over to path based pattern instead. 

Obviously, path based approach is less similar to what a normal environments would look like. E.g. Git servers are accessible at `git.mycompnay.com` or auth services are available at `auth.mycompany.com`. Although it's entirely possible for them to be available under a single domain, it's less common from what I've seen.

nabuskey:
Noticing some strange behaviours in Codespaces. Will troubleshoot next week before merging

jessesanford:
I like the idea of everything squishing into a path based routing sandbox and playing nice together, but we would be creating more pain for ourselves down the road as we try to adopt projects that might be incompatible. At the same time, it may be more of a burden to support both methods for each and every capability we publish. Maybe in the future we shift our default to path based and allow for individual capabilities to use their own hostnames on a case by case basis? It would be up to the capability owner to support the special case then.

nabuskey:
Ok it works with codespaces now. The command to run is something like:

```bash
./idpbuilder create --protocol http  \
  --host ${CODESPACE_NAME}-8080.${GITHUB_CODESPACES_PORT_FORWARDING_DOMAIN} \
  --port 8080 --use-path-routing --ingress-host-name localhost
```

nimakaviani:
super awesome to see this land. thanks for all the hard work!

PR #166, opened 2024-03-04
Enable gitea oci registry

jessesanford:
This is now working with the following patch for gitea: https://github.com/jessesanford/gitea/blob/c2fffabcd5122d258360d4443b6e583fc62e28f1/routers/api/packages/container/container.go#L119-L131

We need to decide if we want to use this forked version temporarily while we shape a PR for upstream or continue to use an additional registry. @nimakaviani @nabuskey @greghaynes wdyt?

### DIFF
--- a/hack/gitea/generate-manifests.sh
+++ b/hack/gitea/generate-manifests.sh
@@ -3,7 +3,7 @@ set -e
 
 INSTALL_YAML="pkg/controllers/localbuild/resources/gitea/k8s/install.yaml"
 GITEA_DIR="./hack/gitea"
-CHART_VERSION="10.1.3"
+CHART_VERSION="10.1.4"
 
 echo "# GITEA INSTALL RESOURCES" >${INSTALL_YAML}
 echo "# This file is auto-generated with 'hack/gitea/generate-manifests.sh'" >>${INSTALL_YAML}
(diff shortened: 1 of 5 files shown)


nabuskey:
We had some discussions around enabling gitea OCI registry.

I think we should enable it by default if users decides to use. My reasoning is that we already install Gitea by default so we can get this feature for "free". I don't see any harm in enabling this feature by default so long as we stay local.

BUT we need to be careful and not rely on this. End users should be able to use whatever registries they want to use. We should not restrict them to using this registry.

nimakaviani:
@jessesanford can you plz drop the link to the change you had to make to gitea for its image registry to work? I am curious if we can find a better way to get around the problem than to have to use a custom image.

jessesanford:
@nabuskey I agree with everything you said. It just so happens this works very well with the ref-implementation workflows I am developing.

@nimakaviani the unofficial PR for my changes is here... it's pretty simple: https://github.com/jessesanford/gitea/pull/1/files

You can also see the discussion of the issue here on the official repo: https://github.com/go-gitea/gitea/issues/29591

I am bullish on this becoming an actual upstream patch as the project has accepted similar contributions for the git repo functionality in the past. See: https://github.com/go-gitea/gitea/issues/19345

tapas4java:
How about adding Harbor as a core application and use as OCI/Image registry? Anyway every platform needs some kind of registry for application images

nimakaviani (approved):
LGTM

PR #167, opened 2024-03-06
Add sleep to see if API server needs a bit

hodgiwabi:


pkg/build/build.go
```diff
@@ -126,6 +126,8 @@ func (b *Build) Run(ctx context.Context, recreateCluster bool) error {
 	}
 
 	setupLog.Info("Adding CRDs to the cluster")
+	// Wait for API Server to come up
+	time.Sleep(15)
 	if err := b.ReconcileCRDs(ctx, kubeClient); err != nil {
 		return err
 	}
```
nimakaviani, line 133: this could work but better would be if we put a retry mechanism in place. what do you think?

hodgiwabi: Sorry this was for some internal testing things--not a real PR for now. If this is something that's needed we'll do a real mechanism not a sleep.

nimakaviani: sounds good.

PR #168, opened 2024-03-08
Enable Plugin scaffolder actions plugin in ref-implementation example

jessesanford:


jessesanford:
@nabuskey See here how I am calling in the new image to enable plugin-scaffolder-actions

PR #169, opened 2024-03-08
re-work ref impl to use path based routing instead

nabuskey:
Updated it to use path routing instead and added support for running it in web browsers through Codespaces. 

During testing, I noticed the default minimum spec given by Codespaces was not enough to run all the examples. I've updated the default Codespaces spec to 4 cores. 

In some environments, the proxy address provided by GitHub may be blocked on host machines due to security configurations provided by organizations. In that case, you cannot run and use this (or any apps) in a web browser. They will need to run it through the GH CLI or locally.


examples/ref-implementation/README.md
```diff
@@ -49,15 +54,15 @@ The only package that cannot be removed this way is Keycloak because other packa
 
 
 #### Accessing UIs
-- Argo CD: https://argocd.cnoe.localtest.me:8443
-- Argo Workflows: https://argo.cnoe.localtest.me:8443
-- Backstage: https://backstage.cnoe.localtest.me:8443
-- Gitea: https://gitea.cnoe.localtest.me:8443
-- Keycloak: https://keycloak.cnoe.localtest.me:8443/admin/master/console/
+- Argo CD: https://cnoe.localtest.me:8443/argocd
+- Argo Workflows: https://cnoe.localtest.me:8443/argo
```
blakeromano, line 58: just for clarity here maybe `/workflows` would be easier to find

csantanapr (approved):
LGTM

examples/ref-implementation/README.md
csantanapr: +1 /workflows or /argo-workflows

greghaynes (approved):

examples/ref-implementation/README.md
nabuskey: renamed to argo-workflows

PR #170, opened 2024-03-09
upgrade controller-gen v0.12.0 -> v0.14.0

nimakaviani:
this fixes the segmentation fault bug on Apple M3 chips when controller-gen generates the crds.

Also formatting of the generated resources is slightly different with the new code-gen tool.


nabuskey (approved):
LGTM

PR #179, opened 2024-03-19
use cnoe/tools-image

jessesanford:
Proposing we build an image rather than install stuff at runtime.

### DIFF
--- a/examples/ref-implementation/keycloak/manifests/keycloak-config.yaml
+++ b/examples/ref-implementation/keycloak/manifests/keycloak-config.yaml
@@ -205,7 +205,7 @@ spec:
             name: config-job
       containers:
         - name: kubectl
-          image: docker.io/library/ubuntu:22.04
+          image: ghcr.io/cnoe-io/tools-image:latest
           volumeMounts:
             - name: keycloak-config
               readOnly: true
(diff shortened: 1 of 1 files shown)


jessesanford:
@nabuskey what do you think about this. I can clean up the commented out debug lines if you think this is a good idea. I don't see why we wouldn't just build an image?

nabuskey:
I didn't want to maintain a docker image so I just used a script to do it. Do we have automations around this?

examples/ref-implementation/keycloak/manifests/keycloak-config.yaml
```diff
@@ -205,7 +205,7 @@ spec:
             name: config-job
       containers:
         - name: kubectl
-          image: docker.io/library/ubuntu:22.04
+          image: ghcr.io/cnoe-io/tools-image:latest
```
cmoulliard, line 208: The github repository which is hosting the code of the container image is empty https://github.com/cnoe-io/tools-image. Is it a mistake ? Do you plan to push the code within the repo ?

cmoulliard (commented):
Don't forget to list within the upstream github repo: cnoe.io/tools-image what the image will include as tools and version !

nimakaviani:
I am also on the side of not wanting to maintain an image for internal tools.

We already added [documentation](https://github.com/cnoe-io/idpbuilder?tab=readme-ov-file#basic-usage) on core packages and their versions. 

I am proposing that this issue can be closed.

nabuskey:
Closing this since we moved our examples to the stacks repo. Please reopen in that repo if this is still relevant.

Issue #165, opened 2024-03-02
Support Kustomize with Custom Packages


greghaynes:
Consider the case where an argo application is in a Kustomize package, and the user would like to create an overlay for local/CI which uses the cnoe:// format.

e.g.

`argocd/resources/application.yaml`:
```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: foo
  namespace: argocd
spec:
  destination:
    server: https://kubernetes.default.svc
  source:
    repoURL: github.com/myapp/config
    path: invalid
    targetRevision: main
```

`argocd/overlays/local`:
```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: dummy
spec:
  destination:
    server: https://kubernetes.default.svc
  source:
    repoURL: cnoe://../../../config/
```

and where `config/` is a directory of k8s resources we desire to install via our argo application.

If we were working with a normal argo application we would be able to specify repoUrl: `cnoe://../../config/` and pass that to idpbuilder, but because we need to run kustomize build beforehand this requires some extra, out-of-band file saving.

PR #180, opened 2024-03-19
custom core packages

nabuskey:
Currently, it's impossible to customize core packages on startup. 

Let's take an example. For a user to configure a very common case of customizing ArgoCD config map to allow for custom health checks:

```yaml
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: argocd-cm
  namespace: argocd
data:
  resource.customizations: |
    argoproj.io/Application:
      health.lua: |
        hs = {}
        hs.status = "Progressing"
        hs.message = ""
        if obj.status ~= nil then
          if obj.status.health ~= nil then
            hs.status = obj.status.health.status
            if obj.status.health.message ~= nil then
              hs.message = obj.status.health.message
            end
          end
        end
        return hs
```
To apply this in idpbuilder, users must:
1. Start idpbuilder and wait for it finish.
2. Go to gitea GUI or clone the argoCD repository.
3. Edit the CM file and push. 

Obviously this workflow is not portable. I think we need to introduce a way to customize core packages.

What I am proposing in this PR is introduction of the `package-custom-file` flag (I don't like this name, please suggest something better). This flag looks something like this:

```shell
$ idpbuilder create --package-custom-file=argocd:pkg/k8s/test-resources/input/argocd-cm.yaml

$ cat pkg/k8s/test-resources/input/argocd-cm.yaml

apiVersion: v1
data:
  application.resourceTrackingMethod: annotation
  resource.exclusions: |
    - kinds:
        - ProviderConfigUsage
      apiGroups:
        - "*"
kind: ConfigMap
metadata:
  labels:
    Test: Data
  name: argocd-cm

$ kubectl get cm argocd-cm -n argocd

apiVersion: v1
data:
  application.resourceTrackingMethod: annotation
  resource.exclusions: |
    - kinds:
        - ProviderConfigUsage
      apiGroups:
        - "*"
kind: ConfigMap
metadata:
  annotations:
    argocd.argoproj.io/tracking-id: argocd:/ConfigMap:argocd/argocd-cm
  creationTimestamp: "2024-03-21T18:03:14Z"
  labels:
    Test: Data
  name: argocd-cm
  namespace: argocd

```

With this flag, you don't have to specify the entire manifest for a package. We will override provided objects (GVK + NS + Name match). E.g. if you just want to change the argocd-cm CM, you can just provide that.

The provided file (`/tmp/cm.yaml` in above example) can be a go template file. Core customization flag values (hostname, protocl, etc) are used to render it before applied to the cluster. 

In addition, you can specify additional objects like more CMs and secrets for ArgoCD to consume. 

I am not sure if we want to support resource deletion initially but I imagine we will need to support it.


jessesanford:
I like the idea, it might be helpful to have a command to print the built in manifests so that folks can understand the defaults and diff between theirs and what's built in.

nabuskey:
> it might be helpful to have a command to print the built in manifests so that folks can understand the defaults and diff between theirs and what's built in. 

agreed. I think we need the export command like we discussed before.

nimakaviani:
this is interesting, thanks for the proposal. 

skimming over the current implementation, looks l ike we will do a full override a given resource from the custom package file. I wonder if we can use an overlay strategy (i.e. Kustomize) to allow folks to add strategic patches instead, so we could change bits in a target file more surgically but also based on some common and already existing patterns. since we have a single file for each of the embedded resources, constructing a base internal should be feasible, I think.

nimakaviani:
my suggestion also goes hand in hand with this proposal https://github.com/cnoe-io/idpbuilder/issues/165

nabuskey:
I don't think kustomize is a good fit. 

1. It requires end users' kustomize knowledge to make changes to core packages.
2. Strategic merge cannot handle all manifest manipulation cases. For example, to remove a field or manipulate arrays, you need other transformers. This means users need to write a full kustomize spec

So if we use kustomize, we are asking the users to: 
1. Get the correct version of manifests from our repo. 
2. Make changes to kustomize so that generated manifests are exactly what they want by running `kustomize build`. 
3. Maintain kustomize transforms. For example, if they want to remove an array item at a specific position, they need to ensure they have not been changed in the base version before they can start using another version of idpbuilder.

I think, in most cases, user made changes to core packages are not that complex. e.g. additional argocd or ingress nignx CMs. While I agree it's a good idea to use an industry standard tools, I am not sure if they are worth it (in terms of implementing such logic and user effort) just to make small changes like these.

cmoulliard:
> Obviously this workflow is not portable. I think we need to introduce a way to customize core packages.

I fully agree. This is today a big issue that we have around the customization of the packages deployed on a cluster as some parameters can be easily configured when we deploy the existing resources (helm, kustomize, etc) but some will require that a user knows how the package A can be customized, etc and the argocd example you provided is a good one ;-) 

Another example could be: How can I create some users/pwd for gitea, repositories, etc

cmoulliard:
> With this flag, you don't have to specify the entire manifest for a package. We will override provided objects (GVK + NS + Name match). E.g. if you just want to change the argocd-cm CM, you can just provide that.

What will happen if the resource already exist ? Will idpbuilder by default merge or replace a CM, secret, etc ? This is an important point that we should carefully log (and why not include as annotations) to inform the users about what it happened

cmoulliard:
What you propose here is inline with my ideas where we should in a future release propose a way for an admin users to define using "profiles/stacks" what they would like to deploy on cluster A or B or C where a profile / stack will be composed of x packages: package A version 1.2.3, package B version 3.4.5, etc and where we will propose a way to customize each package

Remark: Personally and to avoid to execute too many manual commands, I was thinking to use Tekton, Argocd flow or any other flow engine (https://sonataflow.org/serverlessworkflow/latest/index.html) to manage using tasks post installation steps or rollback

cmoulliard:
> It shouldn't touch already existing k8s resources belonging to core packages.

I like the idea about to extend the existing package definition (= what Argocd deploys using the inputs coming from helm chart (or kustomize or etc) instead of merging as that will improve the readability about what idpbuilder do.

Another benefit will also be (according tor the scenario which finally correspond to the users's needs declared in profile/stack), that different configs could be applied against different deployments (cluster a or be or c) using perhaps the same idpbuilder release: 1.2.3 or 1.2.5 or etc

nimakaviani:
> It shouldn't touch already existing k8s resources

ah ok. I was under the impression that people can modify the already existing secrets or CMs as well. if it is only for the first run, then having a flag to pass in the full config makes sense. But why only at create time? in terms of its usability also, only allowing the flag to work at create time will offer a disjoint experience IMO. Basically, your first try with `--create` gets the new config applied, but your followup tries with `--create` wont have an impact, correct?

then what will the cli do in followup runs with both the `--create` and the ` --package-custom-file` flag present? fail? if so, do we expect users to do a `--recreate` in case of any change to the config map? also do they have to remove ` --package-custom-file` after the first run if they want to do some reconciliation as they do the development work?

cmoulliard:
> ah ok. I was under the impression that people can modify the already existing secrets or CMs as well.

We should be very careful here about what we propose:

- If the goal is to change what we install using argocd + helm/kustomize etc then users will certainly complain and tell us: "Why don't you use the secret/cm/etc part of the github repo that we created/customized to create a new IDP builder". Such appraoch is risky and will confuse folks like also increase the debugging process or maintainance.
- If the goal is to apply some "configs" post idpbuilder created, then such option makes sense as it will not hurt what idpbuilder installs or what the user defined as platform to be provisioned. In this case, such additional configs should be part of a "profile/stack" definition (remember my screenshots) and applied against a specific cluster, etc

nabuskey:
During the community meeting we discussed what we want to do with this. 

With the current implementation, modifications to CORE packages are applied at creation only. This was done to ensure we do not break the current behaviour. E.g. We currently do not touch CORE package manifests stored in Gitea repositories after installation. This way users can make modifications in the Gitea UI or clone locally then push without worrying about manifests being overwritten.

Most people in the meeting seem to think we should modify them if specified in the CLI. As a consequence, any modifications made in Gitea UI will be lost. 

I don't have a preference either way. So I'm going to let it modify manifests post creation.

pkg/cmd/create/root.go
```diff
@@ -123,26 +138,41 @@ func validate() error {
...
+	if err != nil {
+		return v1alpha1.PackageCustomization{}, err
+	}
+
+	corePkgs := map[string]struct{}{"argocd": {}, "gitea": {}, "nginx": {}}
+	name := s[0]
+	_, ok := corePkgs[name]
+	if !ok {
+		return v1alpha1.PackageCustomization{}, fmt.Errorf("customization for %s not supported", name)
+	}
```
nimakaviani, line 173:
```suggestion
	if _, ok := corePkgs[name]; !ok {
		return v1alpha1.PackageCustomization{}, fmt.Errorf("customization for %s not supported", name)
	}
```

pkg/cmd/create/root.go
```diff
@@ -123,26 +138,41 @@ func validate() error {
...
+	paths, err := helpers.GetAbsFilePaths([]string{s[1]}, false)
+	if err != nil {
+		return v1alpha1.PackageCustomization{}, err
 	}
 
-	return out, nil
+	err = helpers.ValidateKubernetesYamlFile(paths[0])
+	if err != nil {
+		return v1alpha1.PackageCustomization{}, err
+	}
```
nimakaviani, line 166:
```suggestion
	if err = helpers.ValidateKubernetesYamlFile(paths[0]); err != nil {
		return v1alpha1.PackageCustomization{}, err
	}
```

pkg/cmd/create/root.go
```diff
@@ -123,26 +138,41 @@ func validate() error {
...
+
+	corePkgs := map[string]struct{}{"argocd": {}, "gitea": {}, "nginx": {}}
+	name := s[0]
+	_, ok := corePkgs[name]
+	if !ok {
+		return v1alpha1.PackageCustomization{}, fmt.Errorf("customization for %s not supported", name)
```
nimakaviani, line 172: this logic deserves to be pulled out into a reusable bit

pkg/k8s/deserialize.go
```diff
@@ -51,3 +53,86 @@ func ConvertRawResourcesToObjects(scheme *runtime.Scheme, rawResources [][]byte)
...
+		originals, oErr := kio.FromBytes(originalFile)
+		if oErr != nil {
+			return nil, nil, oErr
+		}
+
+		for j := range originals {
+			id := GetObjectIdentifier(originals[j])
+
+			o, ok := overrideMap[id]
+			if ok {
```
nimakaviani, line 88: can be inlined

nimakaviani (approved):
LGTM. 

One thought I had as was reviewing the PR was, to guard against unintended overrides, I wonder if we should check whether the git repository for the embedded application exists and warn on git repo content getting overwritten before applying the override?

PR #185, opened 2024-03-29
fix recreate flag behavior when using finch

nimakaviani:
- abstracting the runtime provider
- implementing the runtime both for finch and docker
- using the finch cli for container creation
- uses the most recent kind commit with support for detecting the runtime

@nabuskey 
@greghaynes
@estesp 
@csantanapr 

appreciate a review on the above. Also, @estesp curious if there is a more elegant way than using the Finch CLI for the implementation, but ideally without the need to pull in containerd stuff.


estesp:
This looks like a reasonable approach to me

pkg/kind/cluster.go
```diff
@@ -100,13 +103,28 @@ func (c *Cluster) getConfig() ([]byte, error) {
...
+		if rt, err = runtime.NewDockerRuntime(); err != nil {
+			return nil, err
+		}
+	} else if runtime.IsFinchAvailable() {
+		println("Detected runtime: Finch")
+		rt, _ = runtime.NewFinchRuntime()
+	} else {
+		return nil, errors.New("no runtime found")
+	}
+
```
nabuskey, line 124: This bit should go to the runtime pkg. This kind of logic is already available in Kind. I think we should follow their logic: https://github.com/kubernetes-sigs/kind/blob/86c7c2e414662b42fcc36f028a604490bf5489a6/pkg/internal/runtime/runtime.go#L11

This way, users can use w/e runtime with the same configuration they would use for kind.

nimakaviani: sounds good. make the adjustments.

nimakaviani: I realized I didnt fully understand your comment above. will have to do another revisioin. will connect separately

pkg/runtime/util.go
```diff
@@ -0,0 +1,37 @@
+package runtime
+
+import (
+	"errors"
+	"os"
+	"strconv"
+)
+
+func DetectRuntime() (rt IRuntime, err error) {
+	var notFoundErr = errors.New("runtime not found")
```
nabuskey, line 10: Should say runtime not supported or unknown runtime provided.

PR #190, opened 2024-04-10
add one default tag to avoid issues with Ready status

### LINKED ISSUE #189: [Bug]: crossplane demo2 bucket never Ready state
### What is your environment, configuration, and command?

I deployed the ref-implementation with idpbuilder

### What did you do and What did you see instead?

I deploy the template of app plus aws resource (bucket)
The claim and managed resource bucket gets created, the s3 bucket created in aws account, but the bucket status Ready state is "False" never changes to "True"
```
Admin:~/environment/cnoe-examples/hack (main) $ k get objectstorages.awsblueprints.io demotest 
NAME       SYNCED   READY   CONNECTION-SECRET   AGE
demotest   True     False                       19m
Admin:~/environment/cnoe-examples/hack (main) $ k get bucket demotest-fc98v-8ppgr 
NAME                   READY   SYNCED   AGE
demotest-fc98v-8ppgr   False   True     19m
```

### Additional Information. Logs.

```shell
There are not logs in the provider pod to debug this, I created a managed object and started removing fields to spec.

The root cause is tagSet can't be an empty array

apiVersion: s3.aws.crossplane.io/v1beta1
kind: Bucket
metadata:
  name: test-carrlos-2-0410
spec:
    forProvider:
    ....
      tagging:
        tagSet: []

One to fix is for the claim to specify tags, this is what is done in the upstream examples here 
https://github.com/awslabs/crossplane-on-eks/tree/main/examples/aws-provider/composite-resources/s3
```

csantanapr:
Fixes #189

### DIFF
--- a/examples/ref-implementation/crossplane-compositions/manifests/s3/general-purpose.yaml
+++ b/examples/ref-implementation/crossplane-compositions/manifests/s3/general-purpose.yaml
@@ -47,6 +47,10 @@ spec:
               rules:
                 - applyServerSideEncryptionByDefault:
                     sseAlgorithm: AES256
+            tagging:
+              tagSet:
+                - key: cnoe
+                  value: "1"
       patches:
         - type: PatchSet
           patchSetName: common-fields


nabuskey (approved):
LGTM

Issue #150, opened 2024-02-15
Ability to get secrets and configmaps


nabuskey:
When I was porting the reference implementation, I wished there was a way to retrieve all relevant secrets and config map values from the cluster. For example, for the ref implementation, I needed to retrieve values from secrets in two namespaces to log into systems.

What do you think about having a command to get values from secrets and CMs that are labeled in a particular way?

cmoulliard:
> What do you think about having a command to get values from secrets and CMs that are labeled in a particular way?

Why don't we use vault (or alternative) as centralized system to manage for the user(s) their secrets/configs instead of having to query the cluster(s) ?

nabuskey:
fixed in https://github.com/cnoe-io/idpbuilder/pull/193

PR #193, opened 2024-04-15
add get secrets command

nabuskey:
Related to #150 

When you create a new cluster, the first thing you typically do is go to the ArgoCD UI to check on status. We currently tell users to run the kubectl command to get the admin secrets rather than us providing the information directly. This works as long as that's all the information you need from the cluster. 

When you start involving more complex things, it gets messier. For example, when you use the reference implementation example, you have to run a few different kubectl commands to get all credentials required to fully interact with the things deployed in the cluster. I find this quite annoying to document and run as an end user.

I want to allow users to get all information with one command and would like to introduce the `export secret` command. This command does two things:
1. Dump core package credentials. ArgoCD admin password and Gitea admin credentials.
2. Dump any secrets labeled with `cnoe.io/exportable=true`. 

By default it looks like this:
```shell
$ ./idpbuilder export secret
---------------------------
Name: argocd-initial-admin-secret
Namespace: argocd
Data:
  password : abc
---------------------------
Name: gitea-admin-secret
Namespace: gitea
Data:
  password : giteaPassword
  username : giteaAdmin
```

If using the ref impl example, then you just have to label the secrets that you want end users to see.

```shell
$ kubectl label secrets -n keycloak  keycloak-config  cnoe.io/exportable=true
```
```shell
$ ./idpbuilder export secret

---------------------------
Name: argocd-initial-admin-secret
Namespace: argocd
Data:
  password : abc
---------------------------
Name: gitea-admin-secret
Namespace: gitea
Data:
  password : giteaPassword
  username : giteaAdmin
---------------------------
Name: keycloak-config
Namespace: keycloak
Data:
  KC_DB_PASSWORD : abc
  KC_DB_USERNAME : keycloak
  KEYCLOAK_ADMIN_PASSWORD : abc
  POSTGRES_DB : keycloak
  POSTGRES_PASSWORD : abc
  POSTGRES_USER : keycloak
  USER_PASSWORD : abc
```


jessesanford:
I like the idea of using the labels to allow the adhoc credential printing, but will this search all namespaces? If not how do we know in advance which namespaces to query?

cmoulliard:
This is a great idea that you propose here. 2 remarks
- Rename the command `./idpbuilder export secret` to `./idpbuilder list credentials`. Why: credentials implicitly means for a user something related to access a system, service as an admin user, etc
- Use labels: `cnoe.io/core-package=<PACKAGE-NAME>` and `cnoe.io/credentials=true` as every secret (if you are cluster admin) can be "exported" using "kubectl get secret ..."

nimakaviani:
Thanks for this, I like the idea.

big +1 from me on what @cmoulliard suggested. 

As for the command, to stay closer to the kubectl syntax, I would call it `./idpbuilder get credential(s)` with `./idpbuilder get credentials --all or -A` listing all credentials (with an output similar to what you suggested) and also the option to only show credentials or secrets for a single component. for example `./idpbuilder get credential --label cnoe.io/core-package=argo-cd" shows only the following.

```
Name: argocd-initial-admin-secret
Namespace: argocd
Data:
  password : abc
```

The latter is particularly useful for demo purposes with codespaces or c9. Also the `--label` flag should accept an array of labels so you can print several secrets.

cmoulliard:
> `./idpbuilder get credential --label cnoe.io/core-package=argo-cd"

The idp tool should translate that. Command proposed then is: `./idpbuilder get credential <PACKAGE_SHORT_NAME> or ./idpbuilder get credential --package <PACKAGE_SHORT_NAME>` 

Example: `./idpbuilder get credential argocd` or `./idpbuilder get credential --package argocd` or `./idpbuilder get credential --pkg argocd`

**Remark**: If we provision in the future different clusters, then we could extend the command with an additional parameter: `./idpbuilder get credential <PACKAGE_SHORT_NAME> <CLUSTER_NAME>`

pkg/cmd/export/root.go
greghaynes, line 1: I would love at least a smoke test here :)

greghaynes:
Great idea! also +1 to Charles' comments.

nabuskey:
Changed commands to `idpbuilder get secrets`. `credentials` is not really accurate in our context since kubernetes secrets can contain things other than credentials. In addition, if we want to stay close to kubectl command, then we should just do that.

Per Charles' comment, I've introduced two labels keys:
- `cnoe.io/cli-secret` If set to "true", the values are displayed through this command.
- `cnoe.io/package-name`. If set, this values can be used for the `--packages` flag.

I've also introduced a flag called `--packages`. Out of the box, it supports two values: `argocd` and `gitea`. This is really meant for narrowing down secrets that get printed out.

When you run `idpbuilder get secrets`, it gets secrets from core packages and all secrets labeled appropriately.

```sh
./idpbuilder get secrets
---------------------------
Name: gitea-admin-secret
Namespace: gitea
Data:
  password : giteaPassword
  username : giteaAdmin
---------------------------
Name: argocd-initial-admin-secret
Namespace: argocd
Data:
  password : abc
```


```sh
./idpbuilder get secrets -p argocd
---------------------------
Name: argocd-initial-admin-secret
Namespace: argocd
Data:
  password : abc
```

pkg/cmd/get/secrets.go
```diff
@@ -0,0 +1,198 @@
...
+		if pErr != nil {
+			return fmt.Errorf("listing secrets: %w", pErr)
+		}
+
+		for j := range secrets.Items {
+			secretsToPrint = append(secretsToPrint, secretToTemplateData(secrets.Items[j]))
+		}
+	}
+
+	return renderTemplate("templates/secrets.tmpl", outWriter, secretsToPrint)
```
nimakaviani, line 147:
```suggestion
	return renderTemplate(secretTemplatePath, outWriter, secretsToPrint)
```

pkg/cmd/get/secrets.go
```diff
@@ -0,0 +1,198 @@
...
+			continue
+		}
+
+		req, pErr := labels.NewRequirement(v1alpha1.PackageNameLabelKey, selection.Equals, []string{p})
+		if pErr != nil {
+			return fmt.Errorf("building requirement for %s: %w", p, pErr)
+		}
+
+		pkgSelector := selector.Add(*req)
+
```
nimakaviani, line 136: Not sure why this bit needs to sit outside `getSecretsByCNOELabel`. I'd say implement this all in that same function.

nimakaviani (approved):
some minor changes otherwise looks good. 

On a separate note, since we now support selection by package name, I wonder whether there should be another command that lists the packages installed. Also, we should probably pull this k8s specific logic out into a `pkg` repo where they can be added to idpBuilder and later on possible to the cnoe CLI when we need to enable similar functionality for a remote cluster. 

Let me know what you all think and we can create separate issues to refactor things and add the list command. This PR can go in for now IMO.

PR #194, opened 2024-04-18
add labels to expose keycloak secrets

nimakaviani:
add labels to expose ref-implementation credentials when calling `get secrets`


nabuskey (approved):
LGTM

Issue #192, opened 2024-04-11
[Bug]: gitea admin Password not random


csantanapr:
### What is your environment, configuration, and command?

I deployed with idpbuilder create -p examples/ref-implementation

### What did you do and What did you see instead?

The password is generic and stored in gitea git repository and anyone can read it, this causes a problem when I want to expose the URLs outside the linux box (ie ec2 aws) 

### Additional Information. Logs.

```shell
The gitea password should always be random the same way argocd, and keycloak user passwords are random.
```

PR #195, opened 2024-04-18
generate gitea password

nabuskey:
fixes: #192 

Note that this does not in no way makes our gitea installation secure. This is just a bare minimum. You still need to configure Gitea if you want to make it accessible from outside.


nimakaviani (approved):
LGTM

PR #205, opened 2024-04-22
use git providers

nabuskey:
With this PR, I am trying to lay ground work for enabling integrations with GitHub. 

This doesn't change our current behaviours at all. It still uses Gitea as the git provider.


nimakaviani:
Amazing, thanks! will review later today but very excited about this.

cc @greghaynes

pkg/controllers/custompackage/controller.go
```diff
@@ -206,9 +206,13 @@ func (r *Reconciler) reconcileGitRepo(ctx context.Context, resource *v1alpha1.Cu
...
 				Path: absPath,
 			},
-			GitURL:         resource.Spec.GitServerURL,
-			InternalGitURL: resource.Spec.InternalGitServeURL,
-			SecretRef:      resource.Spec.GitServerAuthSecretRef,
+			Provider: v1alpha1.Provider{
+				Name:             v1alpha1.GitProviderGitea,
+				GitURL:           resource.Spec.GitServerURL,
+				InternalGitURL:   resource.Spec.InternalGitServeURL,
+				OrganizationName: v1alpha1.GiteaAdminUserName,
```
blakeromano, line 213: this is a bit confusing here that we have an variable `GiteaAdminUserName` as the `OrganizationName` we probably can have a better variable name or make some clarification here

pkg/controllers/gitrepository/gitea.go
```diff
@@ -0,0 +1,147 @@
...
+	return repoInfo{
+		name:     resp.Name,
+		fullName: resp.FullName,
+		cloneUrl: resp.CloneURL,
+	}, nil
+}
+
+func (g *giteaProvider) getProviderCredentials(ctx context.Context, repo *v1alpha1.GitRepository) (gitProviderCredentials, error) {
+	var secret v1.Secret
+	err := g.Client.Get(ctx, types.NamespacedName{
```
blakeromano, line 55: Maybe I am off basis here but I wonder if we would be better off getting and storing credentials for the git server from a local machine? I can imagine someone having requirements where storing this in K8s secrets would not be allowed.

pkg/controllers/localbuild/controller.go
```diff
@@ -151,7 +151,7 @@ func (r *LocalbuildReconciler) ReconcileArgoAppsWithGitea(ctx context.Context, r
 	}
 	r.shouldShutdown = shutdown
 
-	return ctrl.Result{RequeueAfter: time.Second * 30}, nil
+	return ctrl.Result{RequeueAfter: time.Second * 15}, nil
```
blakeromano, line 154: is there thoughts as to why we changed this? Is there ratelimiting concerns with a lower value?

pkg/controllers/custompackage/controller.go
nabuskey: This is how it's set up right now and this will be changed in the future. The primary purpose of this PR is to keep the current default behaviour but allow for modifications in the future.

pkg/controllers/gitrepository/gitea.go
nabuskey: Great idea. We should support it in the future. Maybe we can discuss it in a feature request issue.

pkg/controllers/localbuild/controller.go
nabuskey: Not really lol. I just wanted to the tests to finish faster. TBH the 30 second  requeue time is also a number I made up when I was writing this first.

pkg/controllers/gitrepository/git_repository.go
```diff
@@ -17,3 +19,30 @@ type GiteaClient interface {
 	SetBasicAuth(username, password string)
 	SetContext(ctx context.Context)
 }
+
+type gitHubClientI interface {
```
nimakaviani, line 23:
```suggestion
type GitHubClient interface {
```

to go with the GiteaClient or we should change the other one to `giteaClient`. Also I dont think we need the `I` for the interface name. not a common practice in this code base.

pkg/controllers/localbuild/controller.go
nimakaviani: I'd say lets revert it back to 30.

pkg/controllers/localbuild/controller.go
```diff
@@ -384,8 +384,12 @@ func (r *LocalbuildReconciler) reconcileGitRepo(ctx context.Context, resource *v
 			Source: v1alpha1.GitRepositorySource{
 				Type: repoType,
 			},
-			GitURL:         resource.Status.Gitea.ExternalURL,
-			InternalGitURL: resource.Status.Gitea.InternalURL,
+			Provider: v1alpha1.Provider{
+				Name:             v1alpha1.GitProviderGitea,
```
nimakaviani, line 388: Given that we are hard-coding it to Gitea, what would be the step for users to choose GitHub instead? Are the users expecetd to add a separate provider to support Github in a way that it co-exists with Gitea? or is there a way to choose alternatives?

pkg/controllers/gitrepository/gitea.go
nimakaviani: @blakeromano mind filing an issue for it?

nimakaviani (commented):
Thank you for this. left a few comments.

Would love to have the README updated as part of this PR. makes the usage flow clearer both for the users and the reviewers.

pkg/controllers/gitrepository/gitea.go
blakeromano: https://github.com/cnoe-io/idpbuilder/issues/237

pkg/controllers/gitrepository/git_repository.go
nabuskey: yeah not sure how it got in there. I think I was messing with it when I was refactoring.

pkg/controllers/localbuild/controller.go
nabuskey: No way to choose an alternative with this PR. To do it, we need to update the CLI and other controllers. Doing it in all one PR will result in a massive PR. Trying to keep a PR small.

nimakaviani (approved):
LGTM, thanks for this!

PR #206, opened 2024-04-23
add git client timeouts

nabuskey:
Fixes an issue where TCP connection to the gitea server times out and the reconciler could not finish its job.

- Adds go-git client timeout values for TCP and HTTP.
- Clone repositories with context instead.


nimakaviani (approved):
LGTM! thanks for the fix

PR #208, opened 2024-04-23
ref-impl: specify content type

nabuskey:
With the ArgoCD upgrade, I guess the token endpoint needs content type specified.


nimakaviani (approved):
LGTM. thanks for the fix

PR #209, opened 2024-04-24
Fix Backstage labels for basic backstage template

mocdaniel:
When deploying the reference implementation and then creating a Backstage component from the `basic` template, the Kubernetes tab would only show the deployment, but no pod information due to the missing `entity-id` label on the deployment's `spec.template.metadata` object.


nabuskey:
Can you please sign your commit to satisfy the DCO check? 

https://stackoverflow.com/questions/1962094/what-is-the-sign-off-feature-in-git-for

https://github.com/apps/dco

mocdaniel:
sure thing, done.

nabuskey (approved):
lgtm

Issue #214, opened 2024-04-25
[Doc] Improve documentation around the different container engines


cmoulliard:
## TODO

- [x] Improve documentation to create a cluster using podman
- [x] Add a table to the README page to list the container engine supported

PR #215, opened 2024-04-25
Improve doc: container engine

cmoulliard:
- Include a prerequisites section covering the container engines. 
- Add a warning and note about podman usage
- See: #214

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,17 @@ This can be useful in several ways:
 * Use within CI to perform integration testing.
 * Use as a local development environment for IDP engineers.
 
+## Prerequisites
+
+A container engine is needed locally such as:
+
+- [Docker desktop](https://www.docker.com/get-started/)
+- [Podman desktop](https://podman-desktop.io/)
+
+**Warning**: idpbuilder can only create a cluster using podman installed as [rootful](https://docs.podman.io/en/latest/markdown/podman-machine-set.1.html#rootful). The rootless mode will raise an error `Error: running kind with rootless provider requires setting systemd property "Delegate=yes", see https://kind.sigs.k8s.io/docs/user/rootless/`.
+
+**Note**: Set the `DOCKER_HOST` env var properly for `podman` to let idpbuilder to talk with the engine (e.g  export DOCKER_HOST="unix:///var/run/docker.sock")
+
 ## Quickstart
 
 ### Download and install the idpbuilder
--- a/README.md
+++ b/README.md
@@ -29,12 +29,13 @@ This can be useful in several ways:
 
 A container engine is needed locally such as:
 
-- [Docker desktop](https://www.docker.com/get-started/)
-- [Podman desktop](https://podman-desktop.io/)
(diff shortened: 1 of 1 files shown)


nabuskey:
Looks good. Can you please fix DCO?

README.md
```diff
@@ -25,6 +25,17 @@ This can be useful in several ways:
...
+## Prerequisites
+
+A container engine is needed locally such as:
+
+- [Docker desktop](https://www.docker.com/get-started/)
+- [Podman desktop](https://podman-desktop.io/)
+
+**Warning**: idpbuilder can only create a cluster using podman installed as [rootful](https://docs.podman.io/en/latest/markdown/podman-machine-set.1.html#rootful). The rootless mode will raise an error `Error: running kind with rootless provider requires setting systemd property "Delegate=yes", see https://kind.sigs.k8s.io/docs/user/rootless/`.
+
+**Note**: Set the `DOCKER_HOST` env var properly for `podman` to let idpbuilder to talk with the engine (e.g  export DOCKER_HOST="unix:///var/run/docker.sock")
```
nabuskey, line 37: We need https://github.com/cnoe-io/idpbuilder/pull/185 merged with proper support for the kind env var instead of overriding `DOCKER_HOST`. @nimakaviani

nimakaviani: I wasnt intending to push #185 for now, since we are waiting for Finch to support Linux. do we want to have it added? Also do you mean that #185 should add support for Podman too?

nabuskey: All I want is for us to respect the `KIND_EXPERIMENTAL_PROVIDER` env var. We don't have to say we support different providers yet. People can experiment with it, if they choose. When they choose to use another provider, we should respect the way kind handles different providers.

In that regard, we should clarify in this PR that the only officially supported kind provider is docker.

cmoulliard: > does `RunsOnPort` work seamlessly with Podman already?

I did successfully a test locally (=> ./idpbuilder create) using podman rootfull

cmoulliard: > In that regard, we should clarify in this PR that the only officially supported kind provider is docker.

As kind works fine now using docker or podman, then I don't see why we should say that only docker desktop is supported as container engine running kind

nabuskey: I guess we haven't really discussed what it means to support different container runtime. At what point do we want to say we do not want to support it? I want us to reach a point where we can say we support these runtimes and here are proof. For that we need way better tests. 

I would love to support different runtime especially podman rootless. 

I think we should talk about this during the community meeting. 

For this PR particularly though, if you want to get it merged fast, then we should say the only officially supported runtime is Docker because that's what we started with and tested against. We can then include the bits about podman rootful but it's not fully tested or supported yet.

cmoulliard: I mentioned now which engine is supported. See last commit: [a19efc7](https://github.com/cnoe-io/idpbuilder/pull/215/commits/a19efc7e5be15ee402539dea83497dffaea25a9f)

nabuskey (approved):
LGTM Thank you!

Issue #184, opened 2024-03-26
Create new git repository for idpbuilder examples and allow the use of URL for flag -p 


csantanapr:
It would be a good idea to create a git repository with examples that are contributed by the community.

This issue covers two main ideas:
1. Create a new github repo under the cnoe-io github org to create a library of examples that work out of the box with one command. This includes to move out the existing [examples](https://github.com/cnoe-io/idpbuilder/tree/main/examples) directory from the idpbuilder git repo
2. Improve the `idpbuilder` to allow to pass an URL to the `-p` flag

I'm not sure what would be the name of the git repo some thing like `cnoe-examples` or `idpbuilder-examples` or just `examples`

This allows for the community to open issues requesting end to end examples, and for contributors to provide examples.

The user experience should that user should be able to run `idpbuilder` with `-p` and point to an url to deploy the example.
```shell
idpbuilder create -p <URL>
```
The URL format would follow kustomize [URL remote targets](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md#remote-targets)
It would look like this
```shell
idpbuilder create -p https://https://github.com/cnoe-io/examples//ref-implementation
open <backstage-url>
```
This would git clone the files to a tmp directory, and start idpbuilder create -p using the tmp dir

Using the kustomize go-lib then we can support ssh git, timeouts, ref/branches, etc.

For now user would need to git clone and use idpbuilder.
```shell
git clone https://https://github.com/cnoe-io/examples
idpbuilder create -p examples/ref-implementation
```

nimakaviani:
this makes sense to me, thanks. I am also wondering whether we should pull out the existing examples directory and submodule it into the core idpbuilder tool for the existing packages to work out of the box.

blakeromano:
This feels like it should tie in with this proposal -> https://github.com/cnoe-io/proposals/issues/1

csantanapr:
@nimakaviani that's what I was proposing, to move out the existing examples directory from this repo to the new examples git repo
>pull out the existing examples directory

nimakaviani:
sure. I am wondering whether submoduling would be a good way to bring it back in or shall we keep them separate

csantanapr:
Friends don't let friends to submodule, unless the friend will automate the upgrade to the submodule reference

nimakaviani:
haha fair!

aatchison:
I agree with the proposal, I was just thinking about this ⬆️

PR #221, opened 2024-04-25
AWS CLI Support to Dev Containers for Backstage Integrations

elamaran11:
This PR is to add AWS CLI Support for Dev Containers . This support is required to be baked in with Dev Containers to support backstage integrations like [terraform backstage integrations](https://github.com/cnoe-io/backstage-terraform-integrations) and others coming soon.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -4,7 +4,8 @@
     "ghcr.io/devcontainers/features/go:1": {
       "version": "1.21"
     },
-    "ghcr.io/devcontainers/features/docker-in-docker:2": {}
+    "ghcr.io/devcontainers/features/docker-in-docker:2": {},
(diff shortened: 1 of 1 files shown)


nabuskey:
Not sure if we want this. It's specific to one example only, and I don't want to introduce something that has nothing to do with the core functionality of idpbuilder to this repo. Maybe we can do it in the packages repo when we move all examples to that.

elamaran11:
@nabuskey Understood but when you say `Maybe we can do it in the packages repo when we move all examples to that` what do we mean by that. You mean to say we should create `.devContainers` in each of our integration repos?

nabuskey:
I was mostly talking about what is discussed here. https://github.com/cnoe-io/idpbuilder/issues/184
I am going to close this PR. Let's talk about this in the issue instead. https://github.com/cnoe-io/idpbuilder/issues/222

PR #233, opened 2024-04-29
feat:add embedded-resources target in Makefile

### LINKED ISSUE #136: Make target for generating embedded resources
Currently, embedded manifests are generated outside of make via scripts in the [hack directory](https://github.com/cnoe-io/idpbuilder/tree/main/hack). 

We should create a new make target that is invoked whenever `make build` is called.

The new target should be called `embedded-resources`

Our Makefile: https://github.com/cnoe-io/idpbuilder/blob/main/Makefile

https://makefiletutorial.com/#getting-started

iron87:
resolve #136

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,8 +63,4 @@ $(ENVTEST): $(LOCALBIN)
 
 .PHONY: embedded-resources
 embedded-resources: 
-	for dir in hack/*; do \
-        if [ -d "$$dir" ]; then \
-            $$dir/generate-manifests.sh; \
-        fi; \
-    done
+	./hack/embedded-resources.sh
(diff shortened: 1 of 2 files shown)


nabuskey (commented):
Thank you very much for this PR! 

I don't think I was specific enough in the issue. That's my bad. Instead of embedding bash scripts in the Makefile, let's create a script in the hack directory so that it's easier to edit in the future. In the script we should specify directory names since we know which directories we care about. Something like this:

`hack/embedded-resources.sh`
```bash
#!/bin/bash

for dir in argo-cd gitea ingress-nginx; do
    ./hack/$dir/generate-manifests.sh;
    if [[ $? -ne 0 ]]; then
        echo "error running script: ./hack/$dir/generate-manifests.sh"
        exit 1
    fi
done

```

Then we can run it like so in Makefile:

```make
.PHONY: embedded-resources
embedded-resources:
	./hack/embedded-resources.sh
```

nabuskey (approved):
LGTM. Thank you very much for this PR!

PR #236, opened 2024-04-30
Fix pull policy

### LINKED ISSUE #235: Leverage local docker image cache
Many of our k8s manifests specify `imagePullPolicy` of always. This includes argocd and gitea which both are sizeable downloads which, as a result of this, are not cached. We should either support setting this as a configuration or change these to `ifNotPresent` so local image caches can be leveraged on slower internet connections.

nimakaviani:
fixes #235

### DIFF
--- a/hack/argo-cd/argocd-application-controller.yaml
+++ b/hack/argo-cd/argocd-application-controller.yaml
@@ -1,0 +1,14 @@
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  labels:
+    app.kubernetes.io/component: application-controller
+    app.kubernetes.io/name: argocd-application-controller
+    app.kubernetes.io/part-of: argocd
+  name: argocd-application-controller
+spec:
+  template:
+    spec:
+      containers:
+        - name: argocd-application-controller
+          imagePullPolicy: IfNotPresent
--- a/hack/argo-cd/argocd-applicationset-controller.yaml
+++ b/hack/argo-cd/argocd-applicationset-controller.yaml
@@ -1,0 +1,14 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app.kubernetes.io/component: applicationset-controller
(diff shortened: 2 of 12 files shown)


greghaynes:
Nice! We should probably open an issue to let users set this as I'm not sure there will be a way for a user to force all of these to update now, e.g. to pull in a new argo release.

greghaynes (approved):

nimakaviani:
it might be slightly easier to do this with `sed` in fact, given that we already do it to alter some other fields in the generated manifests. 

Also to Greg's point, do we want to give the users the option to modify the pull policy on the fly? We dont do it for image references though.

nabuskey:
> it might be slightly easier to do this with sed in fact, given that we already do it to alter some other fields in the generated manifests.
I'd rather stick to kustomize or helm features where possible. Direct manipulation through tools like sed should be for times where it's impossible for us to do it through helm or kusotmize. 

Users can specify w/e manifest they want for core packages right now. We could add a feature just for images though since it's a very common task imo.

blakeromano:
> I'd rather stick to kustomize or helm features where possible. Direct manipulation through tools like sed should be for times where it's impossible for us to do it through helm or kusotmize.

I am of the opinion we should be contributing upstream to add image pull policies rather then trying to do direct manipulation.

nimakaviani:
@blakeromano great point. 

I am going to create an issue with Argo and possibly submit a PR for a fix to their helm chart. In the interim, to unblock us, let's get this merged, and we will do it the right way once Argo CD's helm chart has the ability to merge stuff.

nimakaviani:
asked a question on [the argo cd slack](https://cloud-native.slack.com/archives/C01TSERG0KZ/p1714689772233979).

Alternatively, rather than using the generated install manifest, we can use the argo cd helm chart and set the `imagePullPolicy` to the right value. 

I will wait for a little longer to see if we get a response and if not, we can take the helm chart route.

PR #239, opened 2024-05-01
add initial e2e tests

nabuskey:
This adds initial e2e tests for core packages only.

We are mainly testing rest API responses so we have flexibility in testing tools to use. 

#### Use a scripting languages like Python or JavaScript

- They are easier and less verbose to write. Probably easier for people not familiar with Go to contribute too.
- Rich selection of tools from mature ecosystems.

I think this is one of common approaches. What I do not like about this is the mental overhead when switching to another language. I'm not too good at switching from one language to another when writing things. 
Bringing another language means we are bloating the requirements to work on this project. In addition to installing language runtime, we will need to introduce setup scripts and guides for the chosen language. For example, installing the right version of python then setting up venv.

I felt that the potential repository bloat and overhead were not worth it. So I stuck with Go.

#### Use a testing packages for Go

I thought of using frameworks like https://github.com/gavv/httpexpect. I ultimately chose not to use this because I didn't want to introduce another dependency, but I am open to using something like this as long as we reach a consensus.

#### Tests for examples in the [examples direcotry](./examples)

We should introduce tests for examples like the ref implementation as well, but they shouldn't be part of this repository. We should move them to another repository then test against releases of idpbuilder. Testing methods in the repo can be independent of what we choose here.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -1,0 +1,20 @@
+name: E2E
+
+on:
+  push:
+    branches:
+      - 'main'
+    paths:
+      - '**.go'
+jobs:
+  e2e:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '1.21'
+      - name: Run tests
+        run: |
+          make e2e
--- a/Makefile
+++ b/Makefile
@@ -64,3 +64,7 @@ $(ENVTEST): $(LOCALBIN)
 .PHONY: embedded-resources
 embedded-resources: 
 	./hack/embedded-resources.sh
+
+.PHONY: e2e
+e2e: build
+	go test -v -p 1 --tags=e2e ./tests/e2e/...
--- a/tests/e2e/docker/docker_test.go
+++ b/tests/e2e/docker/docker_test.go
@@ -1,0 +1,111 @@
+//go:build e2e
+
+package docker
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"os"
+	"os/exec"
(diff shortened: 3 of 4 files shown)


blakeromano (commented):
I overall think this is a great start. I wonder if we would have any value in combining this with a tool like chainsaw https://kyverno.github.io/chainsaw/latest/ to do things like make assertions around manifests deployed in the cluster? For example we could make assertions that specific resources have been deployed into the cluster and have particular statuses and use that rather then actually trying to validate that ports are open and making API Calls to ArgoCD?

greghaynes:
I love this! thanks for adding it.

Theres also this: https://github.com/kubernetes-sigs/e2e-framework - I've never used it so not sure if its actually worth it but just FYI.

cmoulliard:
+1 to use `go` as the testing language as this is also the language of this project ;-)
As go testing framework, using an assertion framework will help: https://github.com/stretchr/testify

csantanapr:
I think for idpbuilder scope of function sticking with using go and making so with mininmal external dependency is the way to do, also idpbuilder should not be in charge of testing what's get's deploy to kubernetes other than the core packages which is not that many (argocd, gitea, nginx) so bringing a framework like chainsaw or e2e-framework I would say is too much.

For the examples which ware going to move to another repo, there I can see we use something like e2e-framework or chainsaw to test something like the CNOE ref-implementation. In the new repo we can even support anything that the maintainer of an example folder wants to use, they would add a script (ie. 2e2.sh) that get's call on every folder in examples/stacks provided from github actions to run e2e using the latest version of idpbuilder this way we know that the examples are always healthy with the latest version of idpbuilder

nimakaviani (approved):
+1 to what Manabu and Carlos said on keeping the tests minimal and focused on user functionality.

LGTM

PR #249, opened 2024-05-14
enable serverside apply for tf-controller install

nimakaviani:
Fixes `The ConfigMap is invalid: metadata.annotations: Too long` issue when installing the tf-controller.
Also bumps the version of the tofu controller

The problem is discussed in details here: https://github.com/argoproj/argo-cd/issues/820 and the issue suggests to use `ServerSideApply=true` to skip the use of Kubectl when applying resources.

/cc @elamaran11

### DIFF
--- a/examples/terraform-integrations/tofu-controller.yaml
+++ b/examples/terraform-integrations/tofu-controller.yaml
@@ -11,7 +11,7 @@ spec:
   project: default
   source:
     repoURL: 'https://flux-iac.github.io/tofu-controller'
-    targetRevision: v0.15.0
+    targetRevision: v0.15.1
     helm:
       releaseName: tf-controller
       values: |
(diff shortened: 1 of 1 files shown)


elamaran11:
This issue was intermittent for me and not consistent but thanks for this fix.

nabuskey:
I've run into this issue before too. I think the fix for me was to use the replace option but it comes with some serious drawbacks. if SSA fixes it, I'm all for it.

nimakaviani:
yeah based on all the discussion in the issue I linked you can use eitehr `ServerSideApply` or `Replace`. Given the risks of `Replace` I went with @crenshaw-dev 's recommendation [here](https://github.com/argoproj/argo-cd/issues/820#issuecomment-1636101163).

nabuskey (approved):

PR #256, opened 2024-05-14
feat: GitHub workflow Codespell

vara-bonthu:
**Feature:** Implement Codespell for Automatic Spell Checking

This pull request introduces a new GitHub workflow that leverages the Codespell tool to automatically check for spelling errors across our codebase.

**Benefits:**

- Improved code quality and consistency by identifying potential typos and misspellings.
- Early detection of spelling errors helps maintain a professional codebase.
- Reduced time spent manually checking for spelling mistakes.

**How to Use:**

- No manual intervention is required. Codespell will automatically run on the specified triggers and report any identified spelling errors in the pull request comments.
- You can also run locally with the command `pre-commit run -a` from the root folder. You may have to install pre-commit locally

### DIFF
--- a/.github/workflows/codespell.yaml
+++ b/.github/workflows/codespell.yaml
@@ -1,0 +1,26 @@
+---
+name: Codespell
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+permissions:
+  contents: read
+
+jobs:
+  codespell:
+    name: Check for spelling errors
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Codespell
+        uses: codespell-project/actions-codespell@v2
+        with:
+          check_filenames: true
+          # When using this Action in other repos, the --skip option below can be removed
+          skip: "*.excalidraw,*.git,*.png,*.jpg,*.svg"
--- a/pkg/runtime/finch.go
+++ b/pkg/runtime/finch.go
@@ -11,7 +11,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
-type Conainer struct {
+type Container struct {
 	NetworkSettings NetworkSettings `json:"NetworkSettings"`
 	State           State           `json:"State"`
 }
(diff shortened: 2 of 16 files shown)


pkg/runtime/finch.go
```diff
@@ -11,7 +11,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
-type Conainer struct {
+type Container struct {
```
vara-bonthu, line 14: I can revert this if this is going to cause an issue

pkg/runtime/finch.go
```diff
@@ -59,7 +59,7 @@ func (f *FinchRuntime) ContainerWithPort(ctx context.Context, name string, port
 		return false, err
 	}
 
-	var containers []Conainer
+	var containers []Container
```
vara-bonthu, line 62: same as above

.github/workflows/codespell.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Codespell
+        uses: codespell-project/actions-codespell@v2
+        with:
+          check_filenames: true
+          # When using this Action in other repos, the --skip option below can be removed
+          skip: "*.excalidraw,*.git,*.png,*.jpg,*.svg"
```
nabuskey, line 26: We should skip go.mod and go.sum files.

nabuskey (commented):
I think it's nice that this catches and fixes typos (obviously we have a lot). Is it possible to exclude this check when determining if the PR passes overall checks?

.github/workflows/codespell.yaml
vara-bonthu: done

vara-bonthu:
> I think it's nice that this catches and fixes typos (obviously we have a lot). Is it possible to exclude this check when determining if the PR passes overall checks?

PR checks won't fail with the latest changes

pkg/runtime/finch.go
nabuskey: Looks like these are all contained in this file only so I don't expect to run into any issues.

nabuskey (approved):
We have a lot of typos as shown by this PR and I think it's a good idea to have this. We should have this in the stacks repository as well. Especially when we move the examples directory to it.

nimakaviani (approved):
LGTM

PR #257, opened 2024-05-16
Add localstack as addon package

rattboi:
This adds localstack as an addon to the ref implemention, which involves the following:

    Install localstack as argo application
    Add a new Crossplane ProviderConfig for localstack

I already attempted to add the same directly to the ref impl [here](https://github.com/cnoe-io/idpbuilder/pull/248), but there were lots of cross-cutting concerns there. This is more of a drop-in, although doesn't integrate quite as much.

I think this should be much more acceptable, despite the minor UX downsides.

### DIFF
--- a/examples/localstack-integration/README.md
+++ b/examples/localstack-integration/README.md
@@ -1,0 +1,17 @@
+# Localstack Integration
+
+Please use the below command to deploy an IDP reference implementation with an Argo application that adds Localstack, as well as integrating with Crossplane.
+
+```bash
+idpbuilder create \
+  --use-path-routing \
+  --package-dir examples/ref-implementation \
+  --package-dir examples/localstack-integration
+```
+
+As you see above, this add-on to `idpbuilder` has a dependency on the [reference implementation](../ref-implementation/). This command primarily does the following:
+
+1. Installs `localstack` helmchart as an `argo` application.
+2. Adds localstack crossplane ProviderConfig, targetting localstack
+
+Once the custom package is installed, localstack can be used from the backstage template `app-with-aws-resources`, by changing the `providerConfigName` during the bucket configuration page from `default` to `localstack`.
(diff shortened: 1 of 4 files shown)


nimakaviani:
Thanks for the changes @rattboi ... this looks good to me.

I also played around with the code in your previous PR and landed on something similar but with its full functionality working.

Take a look here: [nimakaviani/idpbuilder/commit/07e92225124ff03a1af113fb94b0cb54b4ea1c1e](https://github.com/nimakaviani/idpbuilder/commit/07e92225124ff03a1af113fb94b0cb54b4ea1c1e)

Assuming that we also `enable-helm` in argocd separately, this should work. 

a few considerations:

- the coredns overlay successfully rewrites the cm-cordns.yaml file and I think a similar pattern can be used to rewrite any k8s resource
- the backstage-template bit is a bit disappointing as it rewrites the git history tree, and we end up losing the other templates. I think we should find a way to have idpBuilder only push to a path to the embedded git repo so we can keep the rest of the git tree untouched. once we have a fix for it, this should also be a minimal change to update the template itself.
- one big caveat is that we are not augmenting but overriding resources, hence quite some duplication. I think for now it is fine, but as we see more addons coming in, we should think about how to fix it.

Anyhow, please take a look and tell me what you think, and we can pull in whatever you like into this PR and merge it.

thanks again for the very interesting line of work.

rattboi:
I took a look at your branch, and it is interesting. 

I think personally I am not a fan of the fact that there is no composability with the overlay you have described, instead doing full swap-outs/replacements of the underlying manifests. I can see that it does work, but it also means that if the base is updated, you need to propagate those changes into whatever addon overlay you have as well. Also, if there were some other addon that needed coredns changes, now you're stuck again, either propagating the localstack changes into your other addon, or you're in an either/or situation...

I think I'm good merging this as-is, and waiting for a composable solution.

PR #258, opened 2024-05-16
fix gitea client timeout

nimakaviani:
we fixed it for the gitCliet in #206 but same can occur with the gitea client and this PR fixes it


nabuskey (approved):
LGTM

PR #261, opened 2024-05-16
appset idpbuilder

csantanapr:
Signed-off-by: Carlos Santana <csantana23@gmail.com>


nimakaviani:
nice to see this coming through!

nabuskey:
Added support for this in https://github.com/cnoe-io/idpbuilder/pull/266
Still need testing but seems ok so far

nabuskey:
Closing this because I think this was just to show how we can use app set.

Issue #93, opened 2023-11-22
Suggestion: Add badges top of README to track activity, job status, etc


cmoulliard:
## Suggestion

I would like to suggest to add top of the README.md file badges in order to track activity such: contributors, activity = commits, jobs

Example of code:
```
<p align="center">
    <a href="https://github.com/halkyonio/primaza-poc/graphs/contributors" alt="Contributors">
        <img src="https://img.shields.io/github/contributors/halkyonio/primaza-poc"/></a>
    <a href="https://github.com/halkyonio/primaza-poc/pulse" alt="Activity">
        <img src="https://img.shields.io/github/commit-activity/m/halkyonio/primaza-poc"/></a>
    <a href="https://github.com/halkyonio/primaza-poc/actions/workflows/push.yml" alt="Build Status">
        <img src="https://github.com/halkyonio/primaza-poc/actions/workflows/push.yml/badge.svg"></a>
</p>
```

![Screenshot 2023-11-22 at 18 10 30](https://github.com/cnoe-io/idpbuilder/assets/463790/2f88cccc-df99-4693-a1dd-0e2d0134d9ec)

xoanmm:
I would like to tackle this issue

nimakaviani:
sounds good @xoanmm ... looking forward to your PR!

xoanmm:
PR created @nimakaviani : https://github.com/cnoe-io/idpbuilder/pull/263

PR #263, opened 2024-05-21
chore: add initial badges

xoanmm:
PR to add Add badges top of README to track activity, job status, etc to solve issue #93


nimakaviani:
Thanks for this @xoanmm.

I think the badges that deliver most value are the following. Can you please update the PR so we have only these ones included? thanks!

- Codespell 
- E2E 
- Go Report Card
- Commit Activity

xoanmm:
Hello @nimakaviani 
I have updated the PR with the requested badges :)

PR #255, opened 2024-05-14
Allow for fetching packages from remote git repositories

nabuskey:
This allows for fetching packages from remote git repositories.

The URL format follows kustomize remote url format.

fixes: #184 

example:
```
idpbuilder create  --package-dir https://github.com/cnoe-io/idpbuilder//examples/basic/package1?version=v0.2.0
```

This should also reduce the number of git clone operations against local and remote repositories.

PR #266, opened 2024-05-29
support app set

nabuskey:
This adds support for application sets. 

Example application sets are located in `pkg/controllers/custompackage/test/resources/customPackages/applicationSet/`

Need:

- [x] #255 merged
- [x] https://github.com/cnoe-io/argocd-api/pull/2 merged


cmoulliard:
README documentation should clearly mention that a package can include `Application CR` and/or `ApplicationSet CR`

api/v1alpha1/custom_package_types.go
```diff
@@ -56,6 +56,7 @@ type ArgoCDPackageSpec struct {
 	ApplicationFile string `json:"applicationFile"`
 	Name            string `json:"name"`
 	Namespace       string `json:"namespace"`
+	Type            string `json:"type"`
```
blakeromano, line 59: Maybe this is beyond the scope of this PR but since these are CRDs and we clearly only support two types we should use the kubebuilder to set enums up for validation. https://book.kubebuilder.io/reference/markers/crd-validation

nabuskey: Oh that's 100% a miss on my part. We should be doing that. I will update it.

pkg/controllers/custompackage/controller.go
```diff
@@ -77,55 +78,101 @@ func (r *Reconciler) reconcileCustomPackage(ctx context.Context, resource *v1alp
...
 				}
+				return ctrl.Result{RequeueAfter: requeueTime}, nil
 			}
-		} else {
-			s := app.Spec.Source
-			res, repo, sErr := r.reconcileArgoCDSource(ctx, resource, s, appName)
+			return ctrl.Result{}, fmt.Errorf("getting argocd application set object: %w", err)
+		}
+
+		foundAppSetObj.Spec = appSet.Spec
```
omrishiv, line 147: the reconcilliation logic is duplicated here with the above case. Is it possible to dedupe?

pkg/controllers/custompackage/controller_test.go
```diff
@@ -38,7 +44,7 @@ func TestReconcileCustomPkg(t *testing.T) {
 		ErrorIfCRDPathMissing: true,
 		Scheme:                s,
 		BinaryAssetsDirectory: filepath.Join("..", "..", "..", "bin", "k8s",
-			fmt.Sprintf("1.27.1-%s-%s", runtime.GOOS, runtime.GOARCH)),
+			fmt.Sprintf("1.29.1-%s-%s", runtime.GOOS, runtime.GOARCH)),
```
omrishiv, line 47: Do you want to pull this out as a constant? It's used again below

omrishiv (commented):
I think this looks good. I think there's an opportunity to clean up the code a bit if possible

pkg/controllers/custompackage/controller_test.go
nabuskey: We need to make this more dynamic and define it at a project level. We need this in make file too. Probably better to just pull from a file or something similar. But I think it's out side scope for this PR.

pkg/controllers/custompackage/controller.go
nabuskey: I considered it but I don't think we'd save any lines of code or performance because we'd have to type cast in the new function then branch out anyways. I figured a bit of duplication here is acceptable.

nimakaviani (approved):
LGTM. thanks for this!

PR #271, opened 2024-06-02
add configuration steps to create a brew release

nimakaviani:
creates the homebrew formula for idpbuilder

Theoretically everything should work but kinda tricky to test this on the actual repo, since we need to cut a release. we can give it a go after it is reviewed and merged.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -20,6 +20,21 @@ builds:
     ignore:
       - goos: linux
         goarch: '386'
+brews:
+  - name: idpbuilder
+    homepage: "https://cnoe.io"
+    repository:
+      owner: nimakaviani
+      name: homebrew-cnoe
+      token: "{{ .Env.HOMEBREW_TOKEN }}"
+    commit_author:
+      name: "GoReleaser"
+      email: "noreply@goreleaser.com"
+    directory: formula
+    install: |
+      bin.install "idpbuilder"
+    test: |
+      system "#{bin}/idpbuilder --version"
 archives:
   - format: tar.gz
     name_template: >-
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -22,6 +22,12 @@ jobs:
       - name: build
         run: |
           OUT_FILE=/tmp/idpbuilder make build
+      - name: Generate a homebrew tap update token
+        id: generate-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ vars.CNOE_HOMEBREW_APP_ID }}
+          private-key: ${{ secrets.CNOE_HOMEBREW_PRIVATE_KEY }}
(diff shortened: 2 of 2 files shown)


.goreleaser.yaml
```diff
@@ -20,6 +20,21 @@ builds:
     ignore:
       - goos: linux
         goarch: '386'
+brews:
+  - name: idpbuilder
+    homepage: "https://cnoe.io"
+    repository:
```
blakeromano, line 26: Do we want the repository for this to be `nimakaviani/homebrew-cnoe` or is this for testing? I would imagine maybe we'd want to consider `cnoe-io/tap` as the repository. It does seem that other big projects like argocd follow the pattern of `project/tap` for example argo uses `argoproj/tap` based on documentation here https://docs.brew.sh/Taps#the-brew-tap-command

nimakaviani: ah, an oversight with the owner name. fixed, thanks!

agreed. tap is a better name. renamed the repo to publish to tap.

blakeromano (approved):
LGTM 🚀

.goreleaser.yaml
```diff
@@ -20,6 +20,21 @@ builds:
...
+      owner: cnoe-io
+      name: homebrew-tap
+      token: "{{ .Env.HOMEBREW_TOKEN }}"
+    commit_author:
```
omrishiv, line 30: I know this is an automated build and we'd want to call that out, but does it make sense to use something more related to CNOE here? Maybe: 
```
name: "CNOEAutomation"
email: "noreply@cnoe.io"
```

omrishiv (commented):
Just a small nit from me. This looks great, thanks!

.goreleaser.yaml
nabuskey: Yea I'd want this to be non-generic.

omrishiv (approved):
Thanks! LGTM 🚀

nabuskey:
Is this still relevant?

cmoulliard:
Adding idpbuilder to homebrew is definitely a plus but then the documentation should be updated

cmoulliard:
LGTM. Approved.

cmoulliard:
Can we go ahead and merge this PR ? Is it possible to test it (= GH Flow) to check what it is produced and to make a test to install it locally ?

nabuskey:
Fix the merge conflict and we can merge imo. As for testing it in a GH workflow, I am not sure tbh. Maybe open an issue and work on it separately? I do like the idea of installing it via home brew.

cmoulliard:
> Maybe open an issue and work on it separately? I do like the idea of installing it via home brew.

As an action should run tonight when nightly tag will be created I will review tomorrow what happened and maybe create a ticket separately

cmoulliard (approved):

PR #276, opened 2024-06-04
remove problematic special chars for bash

nabuskey:
When using idpbuilder with bash, the gitea secret could contain special characters that are not user or bash friendly.  Especially in bash scripts, special characters often have meaning. Removing the following characters: `"`, `\`.

Ref: https://cloud-native.slack.com/archives/C05TN9WFN5S/p1717410133645919

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -17,7 +17,7 @@ import (
 const (
 	chars           = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
 	digits          = "0123456789"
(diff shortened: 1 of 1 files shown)


nimakaviani (approved):
LGTM

cmoulliard:
Backstage is complaining too about the `]` symbol ==>
```
YAMLParseError: Unexpected flow-seq-end token in YAML stream: "]" at line 82, column 31:
       password: *e=sUnx)4Eu-2]k$j~cX-]l$,8EZ"4m7tC!?+C9q
```

nimakaviani:
sorry I missed the context on this one when I merged it. Only saw the slack conversation afterwards. Should have let @cmoulliard review it first. Anyway, we will have to get the fix in for the latest issue with Backstage.

Issue #65, opened 2023-10-25
Rename buildName flag


nabuskey:
As pointed out by @cmoulliard in https://github.com/cnoe-io/idpbuilder/issues/42#issuecomment-1779294207, we should change it. 

Let's rename the `buildName` flag to `name`. This should correspond to the name of the kind cluster. The default name of `localdev` is okay with me unless others would like it changed.

xoanmm:
Created PR #277 to solve the mentioned issue

squidboylan:
Looks like this can be closed? I see `buildName` is deprecated in favor of `name`

PR #277, opened 2024-06-04
chore: rename buildName flag to name

xoanmm:
Solves issue #65

### DIFF
--- a/pkg/cmd/create/root.go
+++ b/pkg/cmd/create/root.go
@@ -20,7 +20,7 @@ import (
 var (
 	// Flags
 	recreateCluster           bool
-	buildName                 string
+	name                      string
 	kubeVersion               string
 	extraPortsMapping         string
 	kindConfigPath            string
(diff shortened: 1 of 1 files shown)


nabuskey:
> I wonder whether we should keep both flags for now with buildName called out as deprecated, and we remove it in a v0.6.0 release? dont want to break people, in case anyone is using it in a CI or something.

Agree with this. The issue was written when we didn't have to worry about backward compatibility just yet. 

Thanks for this PR and would love to merge it once the flag is preserved.

xoanmm:
@nabuskey @nimakaviani so you suggest to modify the actual PR with both flags to keep backward compatibility showing a message like `deprecated, will be removed in future versions use name` in `buildName` flag?

nabuskey:
> @nabuskey @nimakaviani so you suggest to modify the actual PR with both flags to keep backward compatibility showing a message like `deprecated, will be removed in future versions use name` in `buildName` flag?

Yep. It should accept either one, `name` or `build-name`.  If `build-name` is used, print out a message like you said, then continue. Something like `--build-name flag is deprecated and will be removed in a future release. Please use --name instead.`

nabuskey:

cmoulliard:
> Closing this due to inactivity. Feel free to reopen.

I will submit a PR soon ;-)

cmoulliard:
Done through a different PR

Issue #283, opened 2024-06-06
Backstage fails to parse the gitea password: Unexpected flow-seq-end token in YAML stream: "]"


cmoulliard:
## Issue

Backstage fails to parse the gitea password containing some special chars and raises such an error within the backend log `Unexpected flow-seq-end token in YAML stream: "]"`
```
YAMLParseError: Unexpected flow-seq-end token in YAML stream: "]" at line 82, column 31:
       password: *e=sUnx)4Eu-2]k$j~cX-]l$,8EZ"4m7tC!?+C9q
```

**Remark**: The export of the secrets using json output could be an issue too as special chars are converted into unicode chars
```
./idpbuilder get secrets -o json
[
  {
    "name": "gitea-credential",
    "namespace": "gitea",
    "data": {
      "password": "-f!F[-g\u0026V\u0026|\u003e\\7e5[%E7\u003e\\x\u0026(K~,Zjwa3Z8*JHQ~",
      "username": "giteaAdmin"
    }
  },
...
---------------------------
Name: gitea-credential
Namespace: gitea
Data:
  password : -f!F[-g&V&|>\7e5[%E7>\x&(K~,Zjwa3Z8*JHQ~
  username : giteaAdmin
```

_Originally posted by @cmoulliard in https://github.com/cnoe-io/idpbuilder/issues/276#issuecomment-2149072795_

nabuskey:
Hmm I didn't notice the encoding issue in json output. We need to fix.

cmoulliard:
> As for the issue itself, does it work if you use the environment variables in the app-config instead?

No as we still got the ugly password generated containing too many special chars not parseable ;-)

cmoulliard:
The problem is fixed if the backstage user double quotes the gitea password
```
gitea: 
         - host: gitea.cnoe.localtest.me:8443 
           username: giteAdmin
           password: "dhdhdh556DEDSDSD((§!(ç§!ç(§fdn"
```

PR #284, opened 2024-06-06
remove html escape

nabuskey:
As mentioned by @cmoulliard in #283 , the get secret json output escapes certain special characters. This behaviour is not desirable for our purposes.

https://pkg.go.dev/encoding/json#Encoder.SetEscapeHTML


nimakaviani (approved):
LGTM

PR #285, opened 2024-06-06
set e2e timeout to 15m

nabuskey:
The default timeout value of 10m is too short. 

We could execute them in parallel but I fear it may wreck systems with low resources.


nimakaviani (approved):
LGTM

PR #286, opened 2024-06-07
parallel install core

nabuskey:
Currently, we install core packages sequentially and some of them take minimum one minute to install. We can install them in parallel to significantly reduce bootstrap time. 

Current:
```bash
# output of time ./idpbuilder create --recreate
real	2m53.446s
user	0m1.647s
sys	0m0.934s
```

With this PR:
```bash
# output of time ./idpbuilder create --recreate
real	1m27.973s
user	0m1.729s
sys	0m0.919s
```

The absolute amount in reduction varies between systems, but it should reduce the time to bootstrap by a similar percentage.

### DIFF
--- a/pkg/controllers/localbuild/controller.go
+++ b/pkg/controllers/localbuild/controller.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


nimakaviani (approved):
LGTM

PR #288, opened 2024-06-11
ensure only yaml files are processed

nabuskey:
We should be looking at yaml files only when processing custom packages.


nabuskey:
@nimakaviani Fixed.

cmoulliard:
Can you next time please create a ticket otherwise it is needed to look to the issues and/or PR to discover; bug fixed, enhancements, etc @nabuskey @nimakaviani

Issue #282, opened 2024-06-05
Enable on cluster resolution of gitea OCI container registry


jessesanford:
Gitea now has implemented my x-forwarded-for fix for the AppURL resolution in 1.22.0. See Discussion here: https://github.com/go-gitea/gitea/issues/29591 and the fix here: https://github.com/go-gitea/gitea/pull/30891 ... while it is not in the release notes for 1.22 you can see that it was merged into that release branch at https://github.com/go-gitea/gitea/commit/d4c2db39bf239d071c6ac96815efa8fd8b03ad94

Let's bump the gitea version to 1.22+ so that we can make use of the oci registry on cluster.

nimakaviani:
PR https://github.com/cnoe-io/idpbuilder/pull/166 partially fixes this.

As part of fully addressing this issue, we should also provide documentation on using the local registry and potentially making use of it where possible.

cmoulliard:
> As part of fully addressing this issue, we should also provide documentation on using the local registry and potentially making use of it where possible.

Is it dlocumented and where ?

nimakaviani:
yes https://cnoe.io/docs/reference-implementation/installations/idpbuilder/local-oci-registry

cmoulliard:
Using v0.6.0, I got a cert error when I perform a `docker login`
```
docker login -u $GITEA_USERNAME -p $GITEA_PASSWORD $GITEA_URL
WARNING! Using --password via the CLI is insecure. Use --password-stdin.
Error response from daemon: login attempt to gitea.cnoe.localtest.me failed with status: authenticating creds for "gitea.cnoe.localtest.me": pinging container registry gitea.cnoe.localtest.me: Get "https://gitea.cnoe.localtest.me/v2/": dial tcp 127.0.0.1:443: connect: connection refused
```
Such an error will also take place if a an image is build using a Tekton pipeline and pack CLI
```
[pack-builder : pack] Adding buildpack 'paketo-community/ubi-java-buildpack@0.1.1' (diffID=sha256:c3a9faaa3b63a597a3a5f82b1b2d0ea7cf2c0cc08bdc96e2cfbdedcad5c00d91)
[pack-builder : pack] Adding extension 'paketo-community/ubi-java-extension@0.2.0' (diffID=sha256:ba03dfbb2b9d064b5288afbf4cd330347f393cd6586f0c843ce562e80cc17422)
[...]

PR #302, opened 2024-06-14
Docs for #282: added oci registry readme help text

jessesanford:


omrishiv:
Is it worth mentioning how to pull the image from the cluster? I assume we either need a service account with the docker credentials or create a .dockerconfig on the node?

README.md
```diff
@@ -218,6 +219,112 @@ You can also view the updated Application spec by going to this address: https:/
 The second package directory defines two normal ArgoCD applications referencing a remote repository.
 They are applied as-is.
 
+## Local OCI Registry
+
+The local Gitea instance created by idpbuilder contains a built in OCI registry for hosting container images as "packages" in Gitea nomenclature.
+
+It is a standard OCI registry, so the API should be compatible with any tools that are OCI compliant. That includes the `docker` cli.
+
```
csantanapr, line 227: Make a warning here that to login username is case sensitive `giteaAdmin` but when using username/org in OCI image url needs be lower case `giteaadmin` (noticed there are 2 `a`)

I use `giteaAdmin` in docker tag, and lost a lot of time why is not working

README.md
```diff
@@ -218,6 +219,112 @@ You can also view the updated Application spec by going to this address: https:/
...
+For example you can push an image by running:
+
+```bash
+docker login gitea.cnoe.localtest.me:8443
```
csantanapr, line 231: In addition give a one line command example how to login like
```
idpbuilder get secrets -p gitea -o json | jq '.[0].data.password' -r | docker login -u giteaAdmin --password-stdin gitea.cnoe.localtest.me:8443
```

csantanapr (changes requested):

csantanapr:
@omrishiv 
>Awesome, thanks! Is it worth mentioning how to pull the image from the cluster? I assume we either need a service account with the docker credentials or create a .dockerconfig on the node?

no need for pull secrets, the images are public read access

omrishiv:
Ah, thanks for the clarification. Maybe something to think about in the future, but good to know.

README.md
nimakaviani: this is a good call. not sure why we need to call it `giteaAdmin` where `admin` would be good enough. I will create an issue for it so we can change it.

Issue #298, opened 2024-06-14
[Bug]: can't override argocd config settings with idpbuilder create -c argocd:argocd-cm.yaml


csantanapr:
### What is your environment, configuration, and command?

idpbuilder v0.5.0

### What did you do and What did you see instead?

Lets say I want to have different value for argocd cm config like `timeout.reconciliation: 10s` instead of the default 180s

make sure you starting with no cluster
kind delete cluster --name localdev

create argocd-cm.yaml
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: argocd-cm
data:
  application.resourceTrackingMethod: annotation
  resource.exclusions: |
    - kinds:
        - ProviderConfigUsage
      apiGroups:
        - "*"
  timeout.reconciliation: 10s
```

run
```
idpbuilder create -c argocd:argocd-cm.yaml
```
Check the config map has the new override of 10s
```
 k get cm -n argocd argocd-cm -o yaml | grep "timeout.reconciliation: 10s" 
  timeout.reconciliation: 10s
```

Check if environment variable is set on app controllers, as you can see its not set
```
k exec -n argocd argocd-application-controller-0 -- env | grep ARGOCD_RECONCILIATION_TIMEOUT 
<nothing>
```
Restart pod and see that the setting is set.
The configmap with the override is apply after deploying argocd this should be done before.
```
k rollout restart -n argocd sts/argocd-application-controller
k exec -n argocd argocd-application-controller-0 -- env | grep ARGOCD_RECONCILIATION_TIMEOUT 
ARGOCD_RECONCILIATION_TIMEOUT=10s
```

csantanapr:
override for config that controllers fetch dynamically like secrets to connect to helm http/oci or git repos, but not config maps that configure optional environment variables of the controllers

for example in this case here
https://github.com/argoproj/argo-cd/blob/master/manifests/install.yaml#L22780-L22785
```yaml
        - name: ARGOCD_RECONCILIATION_TIMEOUT
          valueFrom:
            configMapKeyRef:
              key: timeout.reconciliation
              name: argocd-cm
              optional: true
```

ricker-flow:
The documentation here appears to be incorrect 
https://github.com/cnoe-io/idpbuilder/blob/main/docs/pluggable-packages.md#runtime-git-server-content-generation

PR #307, opened 2024-06-17
ensure core package customizations are processed

nabuskey:
fixes: #298

One thing to note when overwriting argocd CMs is that you need to have the labels on the CM. Otherwise, ArgoCD will complain about not finding `argocd-cm` CM.

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: argocd-cm
  labels: # these labels need to be present.
    app.kubernetes.io/name: argocd-cm 
    app.kubernetes.io/part-of: argocd
data:
  application.resourceTrackingMethod: annotation
  resource.exclusions: |
    - kinds:
        - ProviderConfigUsage
      apiGroups:
        - "*"
  timeout.reconciliation: 10s
```

@csantanapr ^^


pkg/controllers/resources/idpbuilder.cnoe.io_localbuilds.yaml
```diff
@@ -64,23 +64,22 @@ spec:
...
+                    type: object
+                  packageCustomization:
+                    additionalProperties:
+                      description: PackageCustomization defines how packages are customized
+                      properties:
+                        filePath:
+                          description: FilePath is the absolute file path to a YAML
+                            file that contains Kubernetes manifests.
+                          type: string
+                        name:
```
blakeromano, line 76: Can we make this an enum?

pkg/controllers/resources/idpbuilder.cnoe.io_localbuilds.yaml
```diff
@@ -64,23 +64,22 @@ spec:
...
-                          - name
-                          type: object
-                        type: object
+                    type: object
+                  packageCustomization:
+                    additionalProperties:
+                      description: PackageCustomization defines how packages are customized
+                      properties:
+                        filePath:
+                          description: FilePath is the absolute file path to a YAML
```
blakeromano, line 73: is the plan to not support helm or kustomize?

blakeromano (commented):
I also don't really see much documentation around how this feature is supposed to be invoked/used...

Is this a CLI argument to `idpbuilder create`?

csantanapr:
Good catch on the labels @nabuskey

csantanapr (approved):
/lgtm

nimakaviani (approved):

nabuskey:
> I also don't really see much documentation around how this feature is supposed to be invoked/used...
> 
> Is this a CLI argument to `idpbuilder create`?

We do need to do a better job of documenting these.
See: https://github.com/cnoe-io/idpbuilder/pull/180

Issue #264, opened 2024-05-28
Rename the `-p` flag?


nabuskey:
Do we want to rename the `--package-dir` flag? Right now, it expects a directory that contain ArgoCD application files but this behaviour will likely change soonish. If that happens, we should consider changing this flag's name. 

We should ensure that `--package-dir` flag works going forward for backward compatibility purposes though.

cmoulliard:
> Do we want to rename the `--package-dir` flag?

I'm in favor to rename it in order to been able to create an IDP cluster using packages coming from different locations. Of course, that will imply that we prefix each package with a URI even if we will upload the package resources using a folder/dir 

```
--package files:///a/b/c
--package https://
--package ssh://
--package git://
-- package ftp:///
```

nimakaviani:
yeah should be an easy lift to add the new flag as an alternative and plan to deprecate this in a v0.6 release.

squidboylan:
Is there anything left to do before closing this? Looks like this is fixed

nabuskey:
Agreed. Let's close it.

PR #309, opened 2024-06-18
Add Package Flag and Mark Package-Dir as Deprecated

blakeromano:
Solves: https://github.com/cnoe-io/idpbuilder/issues/264

One thing I didn't know if we wanted to do it in this PR is to setup the skeleton for supporting non-local packages like noted in this comment: https://github.com/cnoe-io/idpbuilder/issues/264#issuecomment-2136865214 I thought that fell out of the scope of the ask in the issue but I am happy to establish that start if it is better to do it in this PR.

### DIFF
--- a/pkg/cmd/create/root.go
+++ b/pkg/cmd/create/root.go
@@ -24,7 +24,9 @@ var (
 	kubeVersion               string
 	extraPortsMapping         string
 	kindConfigPath            string
+	// TODO: Remove extraPackagesDirs after 0.6.0 release
 	extraPackagesDirs         []string
+	extraPackages             []string
 	packageCustomizationFiles []string
 	noExit                    bool
 	protocol                  string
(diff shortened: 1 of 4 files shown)


nimakaviani (changes requested):
This looks great, thank you. There are also a number of places in the README documentation where we refer to the deprecated package. can you please fix them with this PR?

In particular, in both cases where we have `-p` or `--package-dir`, we should rename them to `--package`. 

thank you!

pkg/cmd/create/root.go
```diff
@@ -56,7 +58,10 @@ func init() {
 	CreateCmd.PersistentFlags().StringVar(&protocol, "protocol", "https", "Protocol to use to access web UIs. http or https.")
 	CreateCmd.PersistentFlags().StringVar(&port, "port", "8443", "Port number under which idpBuilder tools are accessible.")
 	CreateCmd.PersistentFlags().BoolVar(&pathRouting, "use-path-routing", false, "When set to true, web UIs are exposed under single domain name.")
+	// TODO: Remove package-dir and deprecation notice after 0.6.0 release
 	CreateCmd.Flags().StringSliceVarP(&extraPackagesDirs, "package-dir", "p", []string{}, "Paths to directories containing custom packages")
+	CreateCmd.Flags().MarkDeprecated("package-dir", "use --package instead")
+	CreateCmd.Flags().StringSliceVar(&extraPackages, "package", []string{}, "Paths to locations containing custom packages")
```
nimakaviani, line 64: lets use the `-p` flag as the shortcut for the `package` flag instead.

blakeromano:
I think I addressed all the PR comments but feel free to let me know if I am missing anything 😄

nimakaviani (approved):
LGTM

PR #313, opened 2024-06-19
Fixes #327: OCI Registry for use with path-based-routing

### LINKED ISSUE #327: Feature: OCI Registry with path based routing
### Have you searched for this feature request?

- [X] I searched but did not find similar requests

### Problem Statement

idpbuilder allows for hosting OCI images on the gitea OCI repo when using subdomains for the capabilities. We should allow for the OCI registry to work with Path based routing.

### Possible Solution

By adding rewrite rules to the gitea nginx ingresses we can make the OCI registry discovery paths accessible

### Alternatives Considered

NA

jessesanford:
Fixes #327 by setting up nginx rewrite rules on the ingress so that the generic OCI registry discovery urls can be found when using path based routing.

### DIFF
--- a/hack/gitea/ingress.yaml.tmpl
+++ b/hack/gitea/ingress.yaml.tmpl
@@ -3,6 +3,70 @@
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
+  name: my-gitea-path-oci-root
+  namespace: gitea
+  annotations:
+    nginx.ingress.kubernetes.io/proxy-body-size: 512m
+spec:
+  ingressClassName: nginx
+  rules:
+    - host: {{ .IngressHost }}
+      http:
+        paths:
+          - backend:
+              service:
+                name: my-gitea-http
+                port:
+                  number: 3000
+            path: /v2
+            pathType: Prefix
+    - host: localhost
+      http:
+        paths:
+          - backend:
+              service:
+                name: my-gitea-http
+                port:
+                  number: 3000
+            path: /v2
+            pathType: Prefix
+---
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: my-gitea-path-oci-repo
+  namespace: gitea
+  annotations:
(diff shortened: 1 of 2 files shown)


jessesanford:
@csantanapr :beers:

cmoulliard:
Can you add a description to this PR plus an issue as this is not clear for the reviewers to figure out what is the purpose of this PR ?

**Remark**: PR should be rejected if they don't include an issue's link and/or description

hack/gitea/ingress.yaml.tmpl
```diff
@@ -2,6 +2,70 @@
 ---
 apiVersion: networking.k8s.io/v1
 kind: Ingress
+metadata:
+  name: my-gitea-path-oci-root
+  namespace: gitea
+  annotations:
+    nginx.ingress.kubernetes.io/proxy-body-size: 512m
```
blakeromano, line 9: Only thought or question on this is 512m big enough if I am pushing and pulling 1GB+ images?

jessesanford:
> Can you add a description to this PR plus an issue as this is not clear for the reviewers to figure out what is the purpose of this PR ?
> 
> **Remark**: PR should be rejected if they don't include an issue's link and/or description

I agree, this is intentionally a quick draft PR to help with a need that @csantanapr had. I will create an issue and appropriate documentation before submitting the actual PR.

hack/gitea/ingress.yaml.tmpl
jessesanford: @nabuskey I think you chose this limit? I agree that many images may have layers > 1GB.

nabuskey: Yes this is a possibility. 512m was chosen for some reason I cannot remember... We can bump it up I think.

blakeromano (approved):

cmoulliard:
> I will create an issue and appropriate documentation before submitting the actual PR.

Don't forget please to add the ticket and description

Issue #137, opened 2024-01-26
Ingress-nginx certificate handling


nabuskey:
This is related to https://github.com/cnoe-io/idpbuilder/issues/127 but not specific to Backstage.

Problem:

Currently, we rely on the default TLS certificate baked in the ingress-nginx controller image to establish TLS sessions for all incoming connections. This works for accessing web UIs from browsers as long as they click on the accept risk button.

This is not so easy for applications. When we want to point an application to the https port on the ingress-nginx service, it will raise an TLS error because it's self-signed and SANs will not match. 
To overcome this kind of TLS issues, many applications provide an option to skip TLS verification steps altogether. This is obviously not a good practice. 

A better approach would be to provide the CA cert to applications so that they can use it to verify certs. 
Unfortunately, this approach is currently impossible in idpbuilder because we rely on the default ingress-nginx certificate. Even if we extract the certificate and make it available for applications to verify against, it will fail because subject name does not match our target names (`*.cnoe.localtest.me`). 

We need to generate our own cert, specify it in the ingress-nginx configuration, then make it available somewhere in the cluster. 

Possible solutions:

Use cert manager to generate self-signed cert for ingress-nginx to use. This means we have to embed manifests like we do now for core packages. This also means we will have a pod essentially doing nothing after the initial install.

Generate it in idpbuilder CLI. Since this is a one-time operation, we could generate it in CLI for ingress-nginx to use.

greghaynes:
Including cert management makes sense to me at a high level. A few thoughts:

- Is there a way to make a clear interface here? Cert management is something I would expect to already be in place for most orgs so its likely this part will need to be plugged out.
- Are we only thinking this would be useful for in-cluster traffic? If we also want it usable for ingress traffic to the cluster, how do we see that workflow working where the user can begin trusting the self signed cert OR they can provide a CA for use?

jessesanford:
+1 for pluggable. As an alternate to self-signed solutions, we can use cert manager with lets encrypt, but folks will need to register their own domains and manage the challenge. 

We can also use minica to stand up a local ca to generate the leaf certs and offer minica along with the rest of the CNOE stack.

Then folks would just have to make sure they include the minica root into their ca bundles, I imagine the pub root cert can be loaded into a ConfigMap or Secret for consumption.

We could sign *.cnoe.localtest.me domains without issue for any services folks using idpbuilder are developing.

[...]

Issue #300, opened 2024-06-14
Feature: CoreDNS configuration for core packages


nabuskey:
### Have you searched for this feature request?

- [X] I searched but did not find similar requests

### Problem Statement

When interacting with core packages like argocd and gitea, the in-cluster DNS name and external (local machine) DNS name may not match out of the box. We work around this by creating a CoreDNS configuration https://github.com/cnoe-io/stacks/blob/main/ref-implementation/coredns/manifests/cm-coredns.yaml in ref impl. 

I don't think we want to do this in external packages. I'd rather take care of these kind of things in idpbuilder itself.

### Possible Solution

Once the cluster is up, we should: 
1. Update the coreDNS CM to use import. https://coredns.io/plugins/import/
2. Enter name resolution information in the referenced file.
3. Wait until name resolves correctly.
This is a one-time operation. We should not do this in case another package overwrites the CM.

cmoulliard:
> 2\. Enter name resolution information in the referenced file.

According to the doc page, the rewrite [rule](https://coredns.io/plugins/rewrite/) should include FROM -> TO but in your example
```
        rewrite name cnoe.localtest.me ingress-nginx-controller.ingress-nginx.svc.cluster.local
```
the incoming names to be resolved including `cnoe.localtest.me` will be rewritten to ` ingress-nginx-controller.ingress-nginx.svc.cluster.local`. Are you sure that the rule is correct ?

cmoulliard:
Is there a way to easily test the rewrite rule post update of the codedns CM ?

cmoulliard:
I did a test where I patched the coreDNS CM 
```
Data
====
Corefile:
----
.:53 {
    errors
    health {
       lameduck 5s
    }
    ready
    rewrite name cnoe.localtest.me ingress-nginx-controller.ingress-nginx.svc.cluster.local
    kubernetes cluster.local in-addr.arpa ip6.arpa {
       pods insecure
       fallthrough in-addr.arpa ip6.arpa
       ttl 30
    }
    prometheus :9153
    forward . /etc/resolv.conf {
       max_concurrent 1000
    }
    cache 30
    loop
    reload
    loadbalance
}
```

of my idpbuilder cluster to include the rewrite rule, restarted the coreDNS deployment and argocd server still complain with the same DNS resolution error when it must create an Application for the git repository -> `http://gitea.cnoe.localtest.me:8443/qshift/my-quarkus-app-job.git`
```
[...]

Issue #293, opened 2024-06-12
backstage cannot access non self signed https server as https://argocd.cnoe.localtest.me:8443


cmoulliard:
## Issue

backstage cannot access non self signed https server as `https://argocd.cnoe.localtest.me:8443` and will raise this error when we scaffold a template using as action `[argocd](argocd:create-resources)` and will report this error: `request to https://argocd.cnoe.localtest.me:8443/api/v1/session failed, reason: self-signed certificate`

![Screenshot 2024-06-12 at 09 50 05](https://github.com/cnoe-io/idpbuilder/assets/463790/2c265096-6f5c-4c42-a8ad-59a6b30780ad)

cmoulliard:
There is a trick which is to export from the secret the `tls.crt` file and to set the following NODE_EXTRA_CA_CERTS env var 
```
kubectl -o json -n argocd get secret/argocd-secret | jq -r '.data."tls.crt"' | base64 -d > tls.crt
export NODE_EXTRA_CA_CERTS=/path/to/argocd/tls.crt
```
but then we will got another error from backstage as the certificate do not include `argocd.cnoe.localtest.me`
```
request to https://argocd.cnoe.localtest.me:8443/api/v1/session failed, 
reason: Hostname/IP does not match certificate's altnames
: Host: argocd.cnoe.localtest.me. is not in the cert's altnames
: DNS:localhost, DNS:argocd-server, DNS:argocd-server.argocd, DNS:argocd-server.argocd.svc, DNS:argocd-server.argocd.svc.cluster.local
```

cmoulliard:
As this problem is fixed using a self certificate and CoreDNS rewrite rule (#317 and #316), I will then close it as I did a test manually

PR #316, opened 2024-06-19
issue self signed certificate 

nabuskey:
fixes: #137 
related to: #300 #293

With this PR, idpbuilder will:
1. Create a self signed certificate.
2. Create a TLS secret in the ingress-nginx NS.
3. Use it as the default TLS certificate for ingress-nginx. (you can still use other certs configured at ingress level)
4. Update ArgoCD's CM to trust the CA.
5. Create a CM that contains cert to the default NS.

I thought about using cert-manager but decided not to use it. For our purposes, we just need a certificate for ingress-nginx for in-cluster and incoming traffic only. Introducing cert-manager means:

1. We need to maintain cert-manager manifests.
2. Slower installation speed. We now need to wait for cert-manager to do its thing and be ready before anything else can continue.
3. Introduces three pods that will do nothing for the most part.

### DIFF
--- a/hack/argo-cd/argocd-tls-certs-cm.yaml.tmpl
+++ b/hack/argo-cd/argocd-tls-certs-cm.yaml.tmpl
@@ -1,0 +1,12 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: argocd-tls-certs-cm
+  labels:
+    app.kubernetes.io/name: argocd-tls-certs-cm
+    app.kubernetes.io/part-of: argocd
+data:
+  'gitea.cnoe.localtest.me': |
+    {{ .SelfSignedCert | indentNewLines 4 }}
+  '{{.Host}}': |
+    {{ .SelfSignedCert | indentNewLines 4 }}
--- a/pkg/build/tls.go
+++ b/pkg/build/tls.go
@@ -28,17 +28,21 @@ const (
 	certificateOrgName = "cnoe.io"
 )
 
-func createIngressCertificateConfigMap(ctx context.Context, kubeClient client.Client, cert []byte) error {
-	cm := &corev1.ConfigMap{
+var (
+	certificateValidLength = time.Hour * 8766 // one year
+)
+
+func createIngressCertificateSecret(ctx context.Context, kubeClient client.Client, cert []byte) error {
+	secret := &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      globals.SelfSignedCertCMName,
 			Namespace: corev1.NamespaceDefault,
 		},
-		Data: map[string]string{
-			globals.SelfSignedCertCMKeyName: string(cert),
+		Data: map[string][]byte{
+			globals.SelfSignedCertCMKeyName: cert,
 		},
 	}
-	err := kubeClient.Create(ctx, cm)
+	err := kubeClient.Create(ctx, secret)
 	if err != nil {
 		if k8serrors.IsAlreadyExists(err) {
 			return nil
@@ -113,7 +117,7 @@ func createSelfSignedCertificate(sans []string) ([]byte, []byte, error) {
 
 	keyUsage := x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign
 	notBefore := time.Now()
-	notAfter := notBefore.Add(time.Hour * 8766) // one year
+	notAfter := notBefore.Add(certificateValidLength)
 
 	serialNumberLimit := new(big.Int).Lsh(big.NewInt(1), 128)
 	serialNumber, err := rand.Int(rand.Reader, serialNumberLimit)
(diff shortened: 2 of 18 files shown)


hack/argo-cd/argocd-tls-certs-cm.yaml.tmpl
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: argocd-tls-certs-cm
+  labels:
+    app.kubernetes.io/name: argocd-tls-certs-cm
+    app.kubernetes.io/part-of: argocd
+data:
+  'gitea.cnoe.localtest.me': |
+    {{ .SelfSignedCert | indentNewLines 4 }}
+  '{{.Host}}': |
+    {{ .SelfSignedCert | indentNewLines 4 }}
```
cmoulliard, line 12: Why do you include too the same SelSignedCert ? Who is `host` ?

pkg/build/tls.go
```diff
@@ -0,0 +1,200 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	certificateOrgName = "cnoe.io"
+)
+
+func createIngressCertificateConfigMap(ctx context.Context, kubeClient client.Client, cert []byte) error {
+	cm := &corev1.ConfigMap{
```
cmoulliard, line 32: Is there a reason to use a CM and not a secret ?

hack/argo-cd/argocd-tls-certs-cm.yaml.tmpl
nabuskey: When idpbuilder is deployed under a single domain, gitea is exposed under <domain>/gitea. So we need to add CA for the root domain as well.

pkg/build/tls.go
nabuskey: We do use secret for ingress-nginx. Since this function only deals with certificate, not privatekey, I don't think it's necessary to use a secret here. Certificates are usually distributed in CMs. 

Having said that, it may actually make sense to use a secret if we are to use ESO, at least until they add support for CMs.

nimakaviani (commented):
Thanks for this. made a few minor comments. otherwise looks good.

I hear your points on the use of self-signed certs and it makes sense in the context of idpBuidler alone. However, if we try to move towards an AppSet strategy that also happens to apply to external clusters, then I wonder whether the idea of using CertManager has a stronger ground.

If users will have cert-manager installed to their clusters by default, and if the idpBuilder is supposed to pave the path for them to eventually transitioin from a dev / test environment to a prod env, shall we just bite the bullet and pay the extra cost of enabling them with the cert-manager in the test environment too?

that said, given the amount of work put into this, this works as an interim solution. But lets revisit this as we move towards expanding on the deployment strategy.

pkg/cmd/create/root.go
```diff
@@ -53,7 +54,7 @@ func init() {
 	CreateCmd.PersistentFlags().StringVar(&kindConfigPath, "kind-config", "", "Path of the kind config file to be used instead of the default.")
 
 	// in-cluster resources related flags
-	CreateCmd.PersistentFlags().StringVar(&host, "host", "cnoe.localtest.me", "Host name to access resources in this cluster.")
+	CreateCmd.PersistentFlags().StringVar(&host, "host", globals.DefaultHostName, "Host name to access resources in this cluster.")
```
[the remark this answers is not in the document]
nabuskey: I'd rather re-name in another PR. There are examples that use this flag so we should be careful.

pkg/build/tls.go
nabuskey: updated to create secret instead.

nabuskey:
I think I'd rather wait for concrete use cases for cert-manager until we pull it into core. Ready for another round of review.

jessesanford:
I agree on waiting for more use cases before bringing in cert manager. It should be reasonably easy to roll forward to it when the time comes.

cmoulliard:
> I think I'd rather wait for concrete use cases for cert-manager until we pull it into core. Ready for another round of review.

Cert Manager can help to deal with many use cases like:
- Generate Self Signed Certificate for the platform where it runs: https://cert-manager.io/docs/configuration/selfsigned/
- Inject OOTB the TLS Secret within the ingress resources - https://cert-manager.io/docs/usage/ingress/
- Trust Certificates issued for domains: LetsEncrypt, etc - https://cert-manager.io/docs/configuration/acme/
- Distribute certifcates as bundle: https://cert-manager.io/docs/trust/trust-manager/
- Create HTTPS/TLS Cert/Key for runtime applications, Java, etc) - https://github.com/snowdrop/pki?tab=readme-ov-file#create-a-pkcs12-using-cert-manager

nimakaviani (approved):
LGTM. thanks!

PR #317, opened 2024-06-20
configure core dns for internal name resolution

nabuskey:
fixes: #300 

This allows for in-cluster hostname resolutions using CoreDNS.

From a pod in the cluster:

```
root@debug-pod:/# dig cnoe.localtest.me +short
10.96.30.155
root@debug-pod:/# dig abc.cnoe.localtest.me +short
10.96.30.155
```

A few points about CoreDNS I should make clear:
1. CoreDNS file allows you to import files matching a pattern.
2. By default, CoreDNS installation only mounts the corefile key within coredns CM.

Taken together, what we need to do to allow our default customization and possible further customization:
1. Update the coredns CM to use the import plugin, then make it look for files matching `/etc/coredns-configs/*.conf`
2. Populate `/etc/coredns-configs/` with two files (sourced from CMs):
    - One that contains our default configuration. That is to resolve the name given by the host flag and all subdomains to the internal ingress address.
    - One that is empty. This is to allow for other packages to update this file for further DNS configurations. In the absence of this file, other packages must update the coredns deployment.
3. Update the coredns deployment to mount the two CMs.

### DIFF
--- a/pkg/build/templates/coredns/deployment-coredns.yaml
+++ b/pkg/build/templates/coredns/deployment-coredns.yaml
@@ -1,0 +1,126 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    k8s-app: kube-dns
+  name: coredns
+  namespace: kube-system
+spec:
+  progressDeadlineSeconds: 600
+  replicas: 2
+  revisionHistoryLimit: 10
+  selector:
+    matchLabels:
+      k8s-app: kube-dns
+  strategy:
+    rollingUpdate:
+      maxSurge: 25%
+      maxUnavailable: 1
+    type: RollingUpdate
+  template:
+    metadata:
+      creationTimestamp: null
+      labels:
+        k8s-app: kube-dns
+    spec:
+      affinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+            - podAffinityTerm:
+                labelSelector:
+                  matchExpressions:
+                    - key: k8s-app
+                      operator: In
+                      values:
+                        - kube-dns
+                topologyKey: kubernetes.io/hostname
+              weight: 100
+      containers:
+        - args:
+            - -conf
+            - /etc/coredns/Corefile
+          image: registry.k8s.io/coredns/coredns:v1.11.1
+          imagePullPolicy: IfNotPresent
+          livenessProbe:
+            failureThreshold: 5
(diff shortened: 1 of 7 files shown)


pkg/build/templates/coredns/deployment-coredns.yaml
```diff
@@ -0,0 +1,126 @@
+apiVersion: apps/v1
+kind: Deployment
```
cmoulliard, line 2: As kind installs OOTB CoreDNS, why do also install it using idpbuilder ? AFAIK It is only needed to patch the coredns ConfigMap !

nabuskey: I don' think this was clear, we are not installing them. We are just updating the spec around mounting config maps. 

A few points about CoreDNS I should make clear:
1. CoreDNS file allows you to import files matching a pattern.
2. By default, CoreDNS installation only mounts the corefile key within coredns CM.

Taken together, what we need to do to allow our default customization and possible further customization:
1. Update the coredns CM to use the import plugin, then make it look for files matching `/etc/coredns-configs/*.conf`
2. Populate `/etc/coredns-configs/` with two files (sourced from CMs):
    - One that contains our default configuration. That is to resolve the name given by the host flag and all subdomains to the internal ingress address.
    - One that is empty. This is to allow for other packages to update this file for further DNS configurations. In the absence of this file, other packages must update the coredns deployment.
3. Update the coredns deployment to mount the two CMs.

nimakaviani (approved):
LGTM

pkg/cmd/create/root.go
```diff
@@ -19,11 +19,11 @@ import (
...
-	recreateCluster           bool
-	buildName                 string
-	kubeVersion               string
-	extraPortsMapping         string
-	kindConfigPath            string
+	recreateCluster   bool
+	buildName         string
+	kubeVersion       string
+	extraPortsMapping string
+	kindConfigPath    string
```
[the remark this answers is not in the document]
nabuskey: Formatting was missed in a previous commit I think.

pkg/build/templates/coredns/deployment-coredns.yaml
nabuskey: This needs to be documented. It is a common practice I think? AKS does something very similar and other k8s provider do it too. https://learn.microsoft.com/en-us/azure/aks/coredns-custom

PR #324, opened 2024-06-21
update gh actions versions

nabuskey:
checkout-go v4 is close to being deprecated because node16 is deprecated in GH.


nimakaviani (approved):
LGTM

PR #326, opened 2024-06-21
use internally and externally resolvable git url

nabuskey:
Currently argocd applications are synced to gitea repositories by using the internal service URLs like `http://my-gitea-http.gitea.svc.cluster.local:3000`. This PR changes the behaviour to use the now internally resolvable URL instead. Users can now use the URL given in the Gitea UI to configure ArgoCD.


nimakaviani (approved):
LGTM.

PR #330, opened 2024-07-04
clean up nightly tags and add nightly tag to artifact name

omrishiv:


.goreleaser.yaml
```diff
@@ -25,7 +25,7 @@ builds:
 archives:
   - format: tar.gz
     name_template: >-
-      {{ .ProjectName }}-{{ .Os }}-{{ .Arch }}
+      {{ .ProjectName }}-{{ .Os }}-{{- if contains .Tag "nightly" }}{{ .Tag }}-{{ end }}{{ .Arch }}
```
nabuskey, line 28: This is going to break the installation script and the installation steps described in the readme. As of now, the install script and instructions installs nightly but at least works. We need to fix the install script. Why do we want to name assets differently for nightly?

omrishiv: This is only for the archive for the release. It doesn't change the binary name in the release

nabuskey: Sorry I wasn't specific enough. We download assets with curl like this:

`curl -L -o ./idpbuilder.tar.gz "https://github.com/cnoe-io/idpbuilder/releases/download/${version}/idpbuilder-$(uname | awk '{print tolower($0)}')-$(uname -m | sed 's/x86_64/amd64/').tar.gz" `

So changing asset name will break the process. Currently, the command will download the latest release which is the nightly build. This is not what we want but at least it does download the correct asset for a supported platform.

I am working on a fix to update the script to download the latest non-nightly version. Still though, why do we want to change the asset name for nightly only?

omrishiv: I'll let @cmoulliard comment on the specific use case. I took a look at the repos across github. Some of them have a specific `nightly` repo to publish their nightly builds to. I don't know if that's a better pattern. Let me split the tag deletion out of this PR so we can get that merge though

nabuskey:

PR #332, opened 2024-07-05
ensure latest stable version is downloaded

nabuskey:
Right now, the install script and the install instructions downloads the latest nightly release. This is not what we want. We want users to use the latest stable version. 

Currently:

```bash
version=$(curl -Ls -o /dev/null -w %{url_effective} https://github.com/cnoe-io/idpbuilder/releases/latest)
version=${version##*/}
echo $version
v0.6.0-nightly.20240705 
```

With this PR:

```bash
version=$(curl -s https://api.github.com/repos/cnoe-io/idpbuilder/releases | grep tag_name | grep -o -e '"v[0-9].[0-9].[0-9]"' | head -n1 | sed 's/"//g')
echo $version
v0.5.0
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,12 +41,11 @@ The following command can be used as a convenience for installing `idpbuilder`, 
 curl -fsSL https://raw.githubusercontent.com/cnoe-io/idpbuilder/main/hack/install.sh | bash
 ```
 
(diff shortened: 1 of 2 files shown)


omrishiv (approved):
LGTM, thanks!

Issue #295, opened 2024-06-13
Feature: Templating


omrishiv:
### Have you searched for this feature request?

- [X] I searched but did not find similar requests

### Problem Statement

Not sure if this should be in cnoe-io/stacks or here, but I think here makes sense since idpbuilder ultimately deploys the stacks:

I think we need a way to properly template the stacks. Right now, we have stacks that have hardcoded urls for the local cnoe deployment. If we want to think about deploying to remote clusters (cnoe-io/idpbuilder#74), we need a way to template out some of the hardcoded bits. 

### Possible Solution

Two possibilities:
1) Add new functionality to idpbuilder that allows taking in a `values.yaml` or similar to give it the ability to render out some common variables. As idpbuilder is ultimately what applies these stacks, that functionality can be centralized there and the stacks in the repo can be kept generic.

2) Rewrite the stacks as helm charts. Helm already has a lot of functionality for advanced templating beyond kustomize. We could have generic values that would deploy the local stack vs overrides that could deploy to remote environments. As a bonus, we'd also maybe be able to leverage chart dependencies, so we could deploy the reference implementation as a collection of sub charts, or the subcharts themselves independent of others.

nimakaviani:
yeah I really like this proposal and it has been top of mind for me too, thinking about how to enable deployment to remote clusters.

I am not so sure about helm charts though. I think the UX that we have for idpbuilder is pretty solid and we should stay with it. whether or not we enable some way of passing configs or enable some of the optionality via some flags, both are approaches we should explore.

tapas4java:
I would propose to have a 'cue' schema, and then we can validate users 'values' file against it before applying.

nimakaviani:
interesting thoguht and cue has certainly come up a numbre of times. feels a bit out of place though given that everything is just yaml config. But I am open to hearing others' thoughts on this and decide accordingly.

omrishiv:
I'm not saying we ditch idpbuilder for helm. I think we can write some of the stacks as helm charts which would allow us to template them and do some conditional magic depending on the target. We would use idpbuilder to template/deploy. I don't love it, but it's something that's currently available. 

Happy to entertain other options, but I think we need to sort this out.

PR #336, opened 2024-07-08
add templating support

nabuskey:
This adds package templating support for idpbuilder. See #295 

1. Introduces a new flag `-F`. This specifies a yaml file with any additional data.
2. Templates are written with `#{` as the delimiter, instead of the usual `{{`. This is to avoid collision with other templates. I am open to changing this. 
3. Values used by core packages are available to custom packages as well. e.g. Host, Port, etc

`pkg/controllers/custompackage/test/resources/customPackages/testDir/params.yaml`
```yaml
custom:
  value: value1
```

`pkg/controllers/custompackage/test/resources/customPackages/testDir/app1/cm.yaml`
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: config
data:
  test1: "one"
  hostName: #{ .Host }#
  customParam: #{ .custom.value }#
```

Run idpbuilder:
```
/idpbuilder create -p pkg/controllers/custompackage/test/resources/customPackages/testDir \
  -F pkg/controllers/custompackage/test/resources/customPackages/testDir/params.yaml
```

This results in CM applied with the following config:

```yaml
apiVersion: v1
data:
  customParam: value1
  hostName: cnoe.localtest.me
  test1: one
kind: ConfigMap
metadata:
  name: config
  namespace: my-app

```

We probably should add functions from https://github.com/Masterminds/sprig as well because these are very commonly used in K8s templating stuff.

Closes: #295

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -28,7 +28,7 @@ var (
 
 type Build struct {
 	name                 string
-	cfg                  util.CorePackageTemplateConfig
+	cfg                  util.PackageTemplateConfig
 	kindConfigPath       string
 	kubeConfigPath       string
 	kubeVersion          string
@@ -43,7 +43,7 @@ type Build struct {
 
 type NewBuildOptions struct {
 	Name                 string
-	TemplateData         util.CorePackageTemplateConfig
+	TemplateData         util.PackageTemplateConfig
 	KindConfigPath       string
 	KubeConfigPath       string
 	KubeVersion          string
@@ -115,7 +115,7 @@ func (b *Build) GetKubeClient(kubeConfig *rest.Config) (client.Client, error) {
 
 func (b *Build) ReconcileCRDs(ctx context.Context, kubeClient client.Client) error {
 	// Ensure idpbuilder CRDs
-	if err := controllers.EnsureCRDs(ctx, b.scheme, kubeClient, b.cfg); err != nil {
+	if err := controllers.EnsureCRDs(ctx, b.scheme, kubeClient, b.cfg.Data); err != nil {
 		setupLog.Error(err, "Error creating idpbuilder CRDs")
 		return err
 	}
@@ -184,6 +184,7 @@ func (b *Build) Run(ctx context.Context, recreateCluster bool) error {
 		return err
 	}
 	b.cfg.SelfSignedCert = string(cert)
+	b.cfg.Data["SelfSignedCert"] = string(cert)
 
 	setupLog.V(1).Info("Running controllers")
 	if err := b.RunControllers(ctx, mgr, managerExit, dir); err != nil {
--- a/pkg/build/coredns.go
+++ b/pkg/build/coredns.go
@@ -22,7 +22,7 @@ const (
 //go:embed templates
 var templates embed.FS
 
-func setupCoreDNS(ctx context.Context, kubeClient client.Client, scheme *runtime.Scheme, templateData util.CorePackageTemplateConfig) error {
+func setupCoreDNS(ctx context.Context, kubeClient client.Client, scheme *runtime.Scheme, templateData util.PackageTemplateConfig) error {
 	checkCM := &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "coredns-conf-default",
@@ -34,7 +34,7 @@ func setupCoreDNS(ctx context.Context, kubeClient client.Client, scheme *runtime
 		return nil
 	}
 
-	objs, err := k8s.BuildCustomizedObjects("", coreDNSTemplatePath, templates, scheme, templateData)
+	objs, err := k8s.BuildCustomizedObjects("", coreDNSTemplatePath, templates, scheme, templateData.Data)
 	if err != nil {
 		return fmt.Errorf("rendering embedded coredns files: %w", err)
 	}
--- a/pkg/build/tls.go
+++ b/pkg/build/tls.go
@@ -173,7 +173,7 @@ func createSelfSignedCertificate(sans []string) ([]byte, []byte, error) {
 	return certOut, privateKeyOut, nil
 }
 
-func setupSelfSignedCertificate(ctx context.Context, logger logr.Logger, kubeclient client.Client, config util.CorePackageTemplateConfig) ([]byte, error) {
+func setupSelfSignedCertificate(ctx context.Context, logger logr.Logger, kubeclient client.Client, config util.PackageTemplateConfig) ([]byte, error) {
 	if err := k8s.EnsureNamespace(ctx, kubeclient, globals.NginxNamespace); err != nil {
 		return nil, err
 	}
--- a/pkg/cmd/create/root.go
+++ b/pkg/cmd/create/root.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/url"
+	"os"
 	"path/filepath"
 	"strings"
 
@@ -16,6 +17,7 @@ import (
 	"github.com/spf13/cobra"
 	"k8s.io/client-go/util/homedir"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/yaml"
 )
 
 var (
@@ -29,6 +31,7 @@ var (
 	extraPackagesDirs         []string
 	extraPackages             []string
 	packageCustomizationFiles []string
+	packageParamsFile         string
 	noExit                    bool
 	protocol                  string
 	host                      string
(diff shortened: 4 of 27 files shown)


nimakaviani:
love this PR. can't wait for this to be ready for review.

we may get some criticism for using comments as a way to template yaml files, similar to how carvel's ytt was under fire at times. we can brainstorm on the right delimeter but otherwise I love the direction here.

omrishiv:
https://giphy.com/gifs/excitement-7h6qHU0ZLR94Y

omrishiv:
image

nabuskey:
We need this for things that get checked into Gitea as well, not just for Kubernetes manifests. For example, in backstage templates, we currently hard code value like this: https://github.com/cnoe-io/stacks/blob/137a12b8e69067b71622097a0858f14865b6d563/ref-implementation/backstage-templates/entities/basic/skeleton/catalog-info.yaml#L12

These values don't work if you are to deploy this with a custom host or protocol like we do in Codespaces or C9. To work around this, currently we ask users to run a script to replace host, port, etc. This means they can't take advantage of the remote package feature we implemented. e.g.

Instead of 
```
idpbuilder create -p https://github.com....
```
They have to do
```bash
git clone https://github.com....
# run script
some_script.sh
idpbuilder create -p ./<PATH>
```
I think we need this feature.

nabuskey:
More thoughts

We need a way to pass runtime information (hostname, protocol, port, etc) to argocd.
Ideally, we can pass such information through Kubernetes construct like CM or secrets.
Unfortunately, ArgoCD does not support it (Flux does).

Without this, packages need to implement their own way to configure their packages. e.g. the replace.sh script in ref-impl.

Two approaches I can think of.


### Store them in a git repository and reference.

ArgoCD supports reading helm values from another repository. 
As part of idpbuilder create command, we create a repository containing runtime information called `idpbuilder-runtime`.
Then use it in ArgoCD spec.
Full example below.

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
spec:
  sources:
  - repoURL: 'https://prometheus-community.github.io/helm-charts'
    chart: prometheus
    targetRevision: 15.7.1
    helm:
      valueFiles:
      - $values/core/values.yaml
  - repoURL: "https://cnoe.localtest.me:8443/gitea/giteadmin/idpbuilder-runtime.git"
    targetRevision: dev
    ref: values
```

This approach is only available to helm. If a user wants to use manifests or kustomize, this won't work. HOWEVER, I think exposing such information in a repository is a good idea. I think we should also expose these information in CM and/or secrets.

### Templating support

We can also solve this with templating support. Drawback is that you pretty much have to have this everywhere.

```yaml
networking:
  host: #{.host}
  protocol: #{.protocol}
  port: #{.protocol}
```

nabuskey:
Another approach is to use argocd custom plugin as suggested by Omri. 

In the plugin, we can look for specific supported variables, then substitute them with values in secrets or CM. This can be done with a simple envsubst command in bash. This gives us portability. Meaning packages published can be used in idpbuilder as well as non-idpbuilder environments.

This is only supported by ArgoCD, and we will need another solution if we are to support another CD solution in the future. e.g. flux. This also means we will introduce a dependency to an image and possibly maintain it. Not a big issue imo but does add a bit of overhead.

nimakaviani:
Thanks, very interesting summary here!

for this statement:
```
Ideally, we can pass such information through Kubernetes construct like CM or secrets.
Unfortunately, ArgoCD does not support it (Flux does).```

is this something that we can contribute back to Argo CD so that we have consistency between Argo CD and flux? looks like if we land on something that is more generic, we wont corner ourselves into having to write and maintain custom plugins or do the more tricky approach of relying on helm values to cross reference configs.

We have done lots of collaboration with Argo previously and I am curious what's the amount of effort to enable passing configs to ArgoCD via CM or secrets.

**Comment by nabuskey** [comment]
> is this something that we can contribute back to Argo CD so that we have consistency between Argo CD and flux? looks like if we land on something that is more generic, we wont corner ourselves into having to write and maintain custom plugins or do the more tricky approach of relying on helm values to cross reference configs.

This is a fairly often requested feature: https://github.com/argoproj/argo-cd/issues/12060

But solving this doesn't solve our problem. This applies to helm only. Does not work for kustomize or pure manifests. Idpbuilder must support all cases (helm, kustomize, and manifests) for custom packages. 

It really comes down to two questions:
1. How do I pass runtime information (hostname, port, protocol, path routing) to custom packages? 
2. Do we want to support custom parameters for custom packages? Some packages may require parameters from users at runtime.

**Comment by blakeromano** [comment]
I think that the general concept of building our own templates is going to break the transition path from local development to production. I think folks will want to rely on the main open source components (ArgoCD, Backstage, etc;) to do templating and manage it within their system using their standards. I think the problem I have is I don't see folks using this outside of a local development environment. And if they use this it will hinder their ability to transition their setup to a remote cluster, specifically production.

The conversation came up as to if we would use this in a remote cluster and really in my dev cycle if I am using a remote cluster I want it as close to production as I want.

I think really where I think is "Are people using IDPBuilder as their deployment tool or as a tool in a production environment" and I think the answer is no even if we get to a point where it is production ready because they are going to be married to a CI tool managing Git files and such which is really where I think we should rely on the Open Source side to worry about templating.

ArgoCD with Applicationsets supports templating out how the application is made and with Backstage there's ways to use nunjucks to do templating. 

I think the maintenance overhead of this and breaking the local development to production or real-world usage is why I don't see the value in this personally.

**Comment by csantanapr** [comment]
I would prefer to use something native to ArgoCD like ApplicationSets to pass custom data
I would make idpbuilder create the `in-cluster` ArgoCD secret, then use the cluster generator to pull the data into an Application spec, this in turns can pass down the data to a helm chart or kustomize to override.

If not using cluster generator, you could put the data file in git, then use the ApplicationSet git generator to pull in the customer data from plain yaml file seating in git.
 
In the Application CR you would do the overrides with 
```yaml
helm:
  valuesObect
    custom: {{.metadata.annotations.custom}}
```
or 
```yaml
kustomize:
  patches:
  - target:
      kind: Deployment
      name: external-dns
    patch: |-
      - op: add
        path: /spec/template/spec/containers/0/args/3
        value: --txt-owner-id={{.metadata.annotations.custom}}
```

This is an example on how I'm using both generators to pull information from cluster generator and git generator
https://github.com/csantanapr/cnoe-examples/blob/main/examples/gitops-bridge/platform/addons/charts/gitops-bridge/templates/applicationsets.yaml
```yaml
# Source: gitops-bridge/templates/applicationsets.yaml
apiVersion: argoproj.io/v1alpha1
kind: ApplicationSet
metadata:
  name: gatekeeper
  namespace: argocd
spec:
  syncPolicy:
    preserveResourcesOnDeletion: true
  goTemplate: true
  goTemplateOptions:
    - missingkey=error
  generators:

    - matrix:
        generators:
          - clusters:
              selector:
                matchLabels:
                  argocd.argoproj.io/secret-type: cluster
          - git:
              repoURL: '{{.metadata.annotations.addons_repo_url}}'
              revision: HEAD
              files:
                - path: "**/stack{{.metadata.annotations.platform_stack_version}}.yaml"
  template:
    metadata:
      name: addon-gatekeeper
    spec:
      project: default
      sources:
      - repoURL: '{{.metadata.annotations.addons_repo_url}}'
        targetRevision: HEAD
        ref: values
      - chart: '{{.addons.gatekeeper.chart}}'
        repoURL: '{{.addons.gatekeeper.repoUrl}}'
        targetRevision: '{{.addons.gatekeeper.targetRevision}}'
        helm:
          releaseName: '{{.addons.gatekeeper.releaseName}}'
          ignoreMissingValueFiles: true
          valueFiles:
              - $values/values/common/{{.addons.gatekeeper.chart}}/values.yaml
              - $values/values/{{.metadata.labels.environment}}/{{.addons.gatekeeper.chart}}/values.yaml
              - $values/values/clusters/{{.name}}/{{.addons.gatekeeper.chart}}/values.yaml
      destination:
        namespace: '{{.addons.gatekeeper.namespace}}'
        name: '{{.name}}'
      syncPolicy:
        automated:
          prune: false
        retry:
          limit: 100
        syncOptions:
          - CreateNamespace=true
          - ServerSideApply=true  # Big CRDs.
```

cmoulliard:
Before to propose a solution, I think that we should start first to think about the user story we will look like to support top of idpbuilder, next to review the technical solutions available, challenge them (pro/con) and finally take a decision.

It is also important that we clarify what we want to do here as the process to customize an internal package could be different from what we could also propose and support for external packages (= user's packages, etc). 

I also agree with the fact that, when we create a cluster and we deploy the internal packages, then we need a way to customize some of the existing YAML resources as some key(s)/value(s) highly depend on the information like: hostname, port, DNS name, etc ... to simplify the maintenance of the existing code and replace what we coded in some places with a templating engine

nabuskey:

cmoulliard:
To be complete here is the PR (= status is draft) developed by argocd to support the proposition reported here and that we could rename into dynamic application parameters: https://github.com/argoproj/argo-cd/pull/12050

PR #338, opened 2024-07-10
support matrix git generator

nabuskey:
ApplicationSet has a generator called matrix generator: https://argo-cd.readthedocs.io/en/stable/operator-manual/applicationset/Generators-Matrix/

This can embed other generators like the git generator. We need to support it for use cases where parameter values are pulled from git.


nimakaviani (approved):
LGTM

Issue #339, opened 2024-07-19
[Bug]: All occurrences of cnoe:// not getting replace everywhere


csantanapr:
### What is your environment, configuration, and command?

latest version

### What did you do and What did you see instead?

I'm building a helm chart that acts like an app of applicationsets and I need to pass down the repoURL down to the helm chart

I have the following app, but only one occurence of cnoe:// gets replace
```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: addons
  namespace: argocd
spec:
  destination:
    namespace: argocd
    name: "in-cluster"
  source:
    repoURL: cnoe://addons
    path: "charts/gitops-bridge"
    helm:
      valuesObject:
        repoURLGit: cnoe://addons
        repoURLValues: cnoe://addons
        applicationsets:
          argocd:
            enabled: false
          aws_load_balancer_controller:
            enabled: false
          gatekeeper:
            enabled: true
  project: default
  syncPolicy:
    automated:
      selfHeal: true
```
As workaround I have to hardcoded the values like this
```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: addons
  namespace: argocd
spec:
  destination:
    namespace: argocd
    name: "in-cluster"
  source:
    repoURL: cnoe://addons
    path: "charts/gitops-bridge"
    helm:
      valuesObject:
        repoURLGit: https://gitea.cnoe.localtest.me:8443/giteaAdmin/idpbuilder-localdev-addons-addons.git
        repoURLValues: https://gitea.cnoe.localtest.me:8443/giteaAdmin/idpbuilder-localdev-addons-addons.git
        applicationsets:
          argocd:
            enabled: false
          aws_load_balancer_controller:
            enabled: false
          gatekeeper:
            enabled: true
  project: default
  syncPolicy:
    automated:
      selfHeal: true
```

PR #340, opened 2024-07-20
add support for helm object

nabuskey:
fixes: #339 

With this PR we now look at the `helmValueObject` field in ArgoCD applications. 

I ended up implementing a general replacement function that could be used for any field in ArgoCD application spec to replace `cnoe://` with derived git URL. This functionality is not enabled in this PR however. I am not sure if this is something we want to support. 

With this PR, Idpbuilder will look at `repoURL` fields and `helmValueObject` fields only.

### DIFF
--- a/pkg/controllers/custompackage/controller.go
+++ b/pkg/controllers/custompackage/controller.go
@@ -2,6 +2,7 @@ package custompackage
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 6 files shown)


csantanapr:
Thanks @nabuskey I think this fix the issue I posted.

I'm ok not having the functionality to replace everywhere for now. Maybe we can wait to see if someone asks for it, and even then we can put behind a experimental env variable to allow it.

csantanapr (approved):

blakeromano (approved):

PR #350, opened 2024-07-31
add basic retry for http requests

nabuskey:
In GH actions, due to limited compute resources availability, e2e tests tend to fail more often than not. This adds basic retry logic to e2e to make sure there's a better chance of completing e2e.


nimakaviani (approved):
LGTM

PR #352, opened 2024-08-02
use self-signed cert for argocd server

nabuskey:
Currently ArgoCD does not use our cert. This PR changes that. This allows services that want to communicate with ArgoCD to do so without disabling TLS verification all together (still need to import the cert).

```
$ openssl s_client -showcerts -servername argocd.cnoe.localtest.me -connect argocd.cnoe.localtest.me:8443 </dev/null | openssl x509 -text | grep 'Subject Alternative Name' -A 1

            X509v3 Subject Alternative Name:
                DNS:cnoe.localtest.me, DNS:*.cnoe.localtest.me
```

### DIFF
--- a/pkg/build/tls.go
+++ b/pkg/build/tls.go
@@ -25,12 +25,32 @@ import (
 )
 
 const (
-	certificateOrgName = "cnoe.io"
+	certificateOrgName     = "cnoe.io"
+	certificateValidLength = time.Hour * 8766
+	argocdTLSSecretName    = "argocd-server-tls"
 )
 
-var (
-	certificateValidLength = time.Hour * 8766 // one year
-)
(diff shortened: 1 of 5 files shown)


pkg/build/tls.go
```diff
@@ -25,12 +25,32 @@ import (
 )
 
 const (
-	certificateOrgName = "cnoe.io"
+	certificateOrgName     = "cnoe.io"
+	certificateValidLength = time.Hour * 8766
```
jessesanford, line 29: Not really a big deal, but maybe we could shorten this? I feel like folks will not be keeping builder clusters up for more than weeks, but just in case we should flag the need to rotate these carts sooner?

jessesanford (approved):
LGTM

pkg/build/tls.go
nabuskey: Yeah we can shorten it. Will follow up with another PR.

PR #354, opened 2024-08-05
skip refresh if owned by app set

nabuskey:
On exit, we currently refresh ALL ArgoCD Applications. If applications are owned by application sets, we should not refresh them ourselves. Let ApplicationSets refresh instead.


nimakaviani (approved):
LGTM

Issue #259, opened 2024-05-16
Bump go version to 1.22


nabuskey:
To upgrade to k8s 1.30, we need go 1.22
Changes:
- https://github.com/cnoe-io/idpbuilder/blob/1ed542233337eeca121b24a25db5b1e96a722c6d/go.mod#L3
- https://github.com/cnoe-io/idpbuilder/blob/1ed542233337eeca121b24a25db5b1e96a722c6d/.github/workflows/e2e.yaml#L17
- https://github.com/cnoe-io/idpbuilder/blob/1ed542233337eeca121b24a25db5b1e96a722c6d/.github/workflows/pr.yaml#L15
- https://github.com/cnoe-io/idpbuilder/blob/1ed542233337eeca121b24a25db5b1e96a722c6d/.github/workflows/release.yaml#L20

Wonder if we we should store them in repo variables.

AryanSharma9917:
@nabuskey, could you assign this issue to me if no one is working on it? I would like to contribute to it!

nabuskey:
Assigned. Thank you and let me know if you have any questions!

PR #360, opened 2024-08-16
Bump Go version to 1.22

AryanSharma9917:
**Description:**

This pull request addresses Issue #259 by updating the Go version from 1.21 to 1.22. The following changes have been made:

- Updated the Go version in `idpbuilder/go.mod` to 1.22.
- Modified `.github/workflows/e2e.yaml`, `.github/workflows/pr.yaml`, and `.github/workflows/release.yaml` to use Go version 1.22.

These changes are necessary to support the upgrade to Kubernetes 1.30, which requires Go 1.22.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/cnoe-io/idpbuilder
 
-go 1.21.3
+go 1.22
 
 require (
 	code.gitea.io/sdk/gitea v0.16.0
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/cnoe-io/idpbuilder
 
-go 1.22
+go 1.22.0
 
 require (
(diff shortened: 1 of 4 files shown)


AryanSharma9917:
@nabuskey could you please review my PR?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/cnoe-io/idpbuilder
 
-go 1.21.3
+go 1.22
```
nabuskey, line 3: I think this needs to be `1.22.0` because that's the minimum required by K8s v0.30.0+.

nabuskey (changes requested):
Thank you very much for this! I think the go mod file need changes but otherwise looks good.

go.mod
AryanSharma9917: Done..!

nabuskey (approved):

Issue #358, opened 2024-08-15
[Bug]: Validate and ensure the container port we need is available and correct. 


nabuskey:
### What is your environment, configuration, and command?

If you run idpbuilder without the necessary container port defined, it will never finish reconciling because we can't talk to the gitea service. We need to ensure:

- Correct port is exposed EVEN IF user supplied kind config does not specify it.
- Make sure we print it out.

Kind config to reproduce the issue: 

```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
  - role: control-plane
    extraMounts:
      - containerPath: /var/lib/kubelet/config.json
        hostPath: ~/.docker/config.json
    extraPortMappings:
      - containerPort: 31337
        hostPort: 31337
      - containerPort: 31340
        hostPort: 31340
      - containerPort: 31333
        hostPort: 31333
```

### What did you do and What did you see instead?

idpbuilder never finishes reconciling.

Issue #366, opened 2024-08-26
[Bug]: `idpbuilder create` silently fails if extraPortMapping` exists in config but doesn't include 443


abangser:
### What is your environment, configuration, and command?

Environment: Local

Configuration:
```
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
  - role: control-plane
    extraMounts:
      - containerPath: /var/lib/kubelet/config.json
        hostPath: /Users/abbybangser/.docker/config.json
    extraPortMappings:
      - containerPort: 31337
        hostPort: 31337
      # - containerPort: 31340
      #   hostPort: 31340
      # - containerPort: 31333
      #   hostPort: 31333
      # - containerPort: 443
      #   hostPort: 8443
      #   protocol: TCP
```

Command:
```
$ ./idpbuilder create --kind-config ./kind-platform-config.yaml
```

### What did you do and What did you see instead?

I attempted to create a cluster using kind config, specifically config that included extra port mapping. This created an error in Gitea becoming healthy. 

This was solved in slack ([message](https://cloud-native.slack.com/archives/C05TN9WFN5S/p1723763250039119?thread_ts=1723757966.723879&cid=C05TN9WFN5S)) by adding another port mapping to the list, but the user experience for error handling could be improved.

1. The idpbuilder command would not ever exit as is
2. When using `ctrl+c` to exit the command, I get a successful message which is confusing

### Additional Information. Logs.

idpbuilder version:
```
$ ./idpbuilder version
idpbuilder 0.7.0-nightly.20240815 go1.21.3 darwin/arm64
```

Full kind config:
```
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
  - role: control-plane
    extraMounts:
      - containerPath: /var/lib/kubelet/config.json
        hostPath: /Users/abbybangser/.docker/config.json
    extraPortMappings:
      - containerPort: 31337
        hostPort: 31337
      # - containerPort: 31340
      #   hostPort: 31340
      # - containerPort: 31333
      #   hostPort: 31333
      # - containerPort: 443
      #   hostPort: 8443
      #   protocol: TCP
```

Specific log line that showed Gitea failed:
```
[...]

PR #368, opened 2024-08-26
ensure ingress port is available

nabuskey:
fixes: #358, #366

With this we always insert the ingress port we need when a custom kind configuration is provided.


nimakaviani (approved):
LGTM

punkwalker:
@nabuskey,
How about this approach?
Parse Base/default Template and then merge additional config passed by user? Instead of adding default fields to passed config?
This way, we can maintain basic config as well as allow user to control additional configuration?

PR #369, opened 2024-08-27
Updated controller tools version

AryanSharma9917:
## Description
This PR updates the controller-gen version in the `idpbuilder` module to align with the updated Kubernetes version.

## Changes Made
- Updated the `CONTROLLER_TOOLS_VERSION` in the `Makefile` from `v0.14.0` to `v0.15.0`.

## Linked Issue
This PR addresses issue #364 and is related to the Kubernetes update mentioned in issue #260.

PR #371, opened 2024-08-28
run controller gen to update crds

nabuskey:
Looks like controller gen was updated but was not run in #369 and nightly release is failing.


tapas4java (approved):

nimakaviani (approved):
LGTM

Issue #367, opened 2024-08-26
[Bug]: If using `--kind-config` it must include networking config


abangser:
### What is your environment, configuration, and command?

Environment: Local

Configuration:
```
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
  - role: control-plane
    extraMounts:
      - containerPath: /var/lib/kubelet/config.json
        hostPath: /Users/abbybangser/.docker/config.json
```

Command:
```
$ ./idpbuilder create --kind-config ./kind-platform-config.yaml
```

### What did you do and What did you see instead?

I tried to use kind config flag, but in doing so seemed to remove some key features that are auto populated otherwise. Namely the `extraPortMapping` which provides the networking between services and likely the `kubeadmConfigPatches` and `containerdConfigPatches` config is necessary too for the networking.

I would ideally see these default values still used, but even just documenting what is necessary if you do provide your own config could work for now.

### Additional Information. Logs.

idpbuilder version:
```
$ ./idpbuilder version
idpbuilder 0.7.0-nightly.20240815 go1.21.3 darwin/arm64
```

Full kind config:
```
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
  - role: control-plane
    extraMounts:
      - containerPath: /var/lib/kubelet/config.json
        hostPath: /Users/abbybangser/.docker/config.json
```

Top of logs for my run:
```
$ ./idpbuilder create --kind-config ./kind-platform-config.yaml
time=2024-08-26T20:57:59.620+01:00 level=INFO msg="Creating kind cluster" logger=setup
time=2024-08-26T20:57:59.632+01:00 level=INFO msg="Runtime detected" logger=setup provider=docker
########################### Our kind config ############################
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
  - role: control-plane
    image: kindest/node:v1.27.3
    extraMounts:
      - containerPath: /var/lib/kubelet/config.json
        hostPath: /Users/abbybangser/.docker/config.json

#########################   config end    ############################
```

Top of logs for a run *without* kind config:
```
$ ./idpbuilder create --kube-version v1.27.3
time=2024-08-26T20:55:37.619+01:00 level=INFO msg="Creating kind cluster" logger=setup
time=2024-08-26T20:55:37.635+01:00 level=INFO msg="Runtime detected" logger=setup provider=docker
########################### Our kind config ############################
[...]

PR #377, opened 2024-08-29
ensure ingress-nginx node label is applied

nabuskey:
fixes: #366 
fixes: #367


pkg/kind/cluster.go
```diff
@@ -241,28 +246,61 @@ func (c *Cluster) ensureCorrectConfig(in []byte) (kindv1alpha4.Cluster, error) {
...
+					}
+				}
+				break nodes
+			}
+		}
+		if node.Labels != nil {
+			v, ok := node.Labels[ingressNginxNodeLabelKey]
+			if ok && v == ingressNginxNodeLabelValue {
+				appendIngresNodeLabel = false
+				nodePosition = i
```
nimakaviani, line 279: I wonder if this can be further simplified. I am not fond of label breaks in go, but leave it up to you.

nimakaviani (changes requested):
just a minor typo and a comment

pkg/kind/cluster.go
```diff
@@ -241,28 +246,61 @@ func (c *Cluster) ensureCorrectConfig(in []byte) (kindv1alpha4.Cluster, error) {
 	if err != nil {
 		return kindv1alpha4.Cluster{}, fmt.Errorf("parsing kind config: %w", err)
 	}
-
+	// the port and ingress-nginx label must be on the same node to ensure nginx runs on the node with the right port.
 	appendNecessaryPort := true
+	appendIngresNodeLabel := true
```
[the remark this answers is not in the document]
nabuskey: fixed.

pkg/kind/cluster.go
nabuskey: It's really just comes down to last one winning or the first one winning. I chose first one winning. If we do last one wins, we can avoid using label breaks.

nimakaviani: i like first one winning too.

Here is an alternative way of writing it (thanks ChatGPT). I dont think it is any easier tho. let's leave it.

```
for i, node := range parsedCluster.Nodes {
    checkNodeLabels := func() {
        if node.Labels != nil {
            if v, ok := node.Labels[ingressNginxNodeLabelKey]; ok && v == ingressNginxNodeLabelValue {
                appendIngresNodeLabel = false
            }
        }
    }

    for _, pm := range node.ExtraPortMappings {
        if strconv.Itoa(int(pm.HostPort)) == c.cfg.Port {
            appendNecessaryPort = false
            nodePosition = i
            checkNodeLabels()
            break
        }
    }

    if appendNecessaryPort { // If the port was not matched, check labels
        checkNodeLabels()
        if !appendIngresNodeLabel {
            nodePosition = i
            break
        }
    }
}
```

nimakaviani (approved):
LGTM

PR #380, opened 2024-09-01
update default kube version to v1.30

tapas4java:
updating default k8s version to v1.30 for kind cluster


nabuskey:
Thanks for this! Could you please fix DCO?

PR #381, opened 2024-09-03
update default k8s version to v1.30

tapas4java:
Updating default k8s version

### DIFF
--- a/pkg/cmd/create/root.go
+++ b/pkg/cmd/create/root.go
@@ -47,7 +47,7 @@ func init() {
 	// cluster related flags
 	CreateCmd.PersistentFlags().BoolVar(&recreateCluster, "recreate", false, "Delete cluster first if it already exists.")
 	CreateCmd.PersistentFlags().StringVar(&buildName, "build-name", "localdev", "Name for build (Prefix for kind cluster name, pod names, etc).")
-	CreateCmd.PersistentFlags().StringVar(&kubeVersion, "kube-version", "v1.29.2", "Version of the kind kubernetes cluster to create.")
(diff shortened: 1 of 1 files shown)


nabuskey (approved):
LGTM

OneCricketeer:
Should `ENVTEST_K8S_VERSION` be updated as well?

pkg/cmd/create/root.go
```diff
@@ -47,7 +47,7 @@ func init() {
 	// cluster related flags
 	CreateCmd.PersistentFlags().BoolVar(&recreateCluster, "recreate", false, "Delete cluster first if it already exists.")
 	CreateCmd.PersistentFlags().StringVar(&buildName, "build-name", "localdev", "Name for build (Prefix for kind cluster name, pod names, etc).")
-	CreateCmd.PersistentFlags().StringVar(&kubeVersion, "kube-version", "v1.29.2", "Version of the kind kubernetes cluster to create.")
+	CreateCmd.PersistentFlags().StringVar(&kubeVersion, "kube-version", "v1.30.0", "Version of the kind kubernetes cluster to create.")
```
OneCricketeer, line 50: Why not latest patch release? 1.30.4?

cmoulliard:
A table about the kubernetes version(s) supported should be included part of the documentation and release note !

pkg/cmd/create/root.go
nabuskey: Yeah good idea. I looked at the kind version we are using and its default k8s version (`v1.30.0`).  Bumping it up to the next patch release is a good idea. Your PR would be awesome here 😃

nabuskey:
Agreed. I really would love to automate this bit for sure.

PR #383, opened 2024-09-06
Remove replace directives from go.mod

greghaynes:
These got added back when we dependend on argocd directly. They are no longer needed, and they break the ability to use go install command to install idpbuilder.


nabuskey (approved):
LGTM

PR #388, opened 2024-09-16
Added config for gitea to support webhooks trigger

mmaroth:
Gitea webhook event POST call is failing with `webhook can only call allowed HTTP servers (check your webhook.ALLOWED_HOST_LIST setting)`, fixing this issue by adding webhook settings to allow private IPs and Skip TLS Verify.


nabuskey (approved):
LGTM

Gitea supports webhook a few configurations. We use `private` as the default because most of what we do are in-cluster. https://docs.gitea.com/administration/config-cheat-sheet?_highlight=webhook/administration/config-cheat-sheet#webhook-webhook

Skipping TLS verification shouldn't happen but we don't have gitea configured to use the TLS cert yet.

Issue #378, opened 2024-08-29
Feature: Gitea auth support


abangser:
### Have you searched for this feature request?

- [X] I searched but did not find similar requests

### Problem Statement

idpbuilder generates Gitea with a strong password. This password has special characters enabled and can creates issues (such as in #283) when trying to use it in command/other applications.

### Possible Solution

Automatically generate a token for the admin user and store in a known secret. This will allow more secure authentication methods and remove the need to change the password.

### Alternatives Considered

There are a number of options we discussed in Slack with @nabuskey including:
1. Reducing the complexity of the password (not ideal as it reduces the security of gitea if made publicly available)
  a. Allow people to opt into this in some way with a flag
2. Create a git token by default for admin user and store in a known secret

PR #392, opened 2024-09-27
generate gitea token

nabuskey:
This PR makes idpbuilder generate a Gitea admin token. This is useful for use cases such as:

- Cloning Gitea repositories.
- Performing some sort of Gitea related operations. e.g. configure webhook for a repository.

Partially fixes: https://github.com/cnoe-io/idpbuilder/issues/378

Example output: 

```
./idpbuilder get secrets -o json -p gitea
[
  {
    "name": "gitea-credential",
    "namespace": "gitea",
    "data": {
      "password": "E_aaaaa'",
      "token": "aaaaaa",
      "username": "giteaAdmin"
    }
  }
]
```

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -3,12 +3,16 @@ package util
 import (
 	"context"
 	"crypto/rand"
+	"crypto/tls"
 	"fmt"
 	"math"
 	"math/big"
 	mathrand "math/rand"
+	"net"
+	"net/http"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/cnoe-io/idpbuilder/api/v1alpha1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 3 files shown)


cmoulliard:
What is the duration of the `Token` ? Can a user renew it ? @nabuskey

nabuskey:
> What is the duration of the `Token` ? Can a user renew it ? @nabuskey

No expiration. I don't think Gitea supports it until they move to JWT: https://github.com/go-gitea/gitea/issues/4793

cmoulliard (commented):
Could also mention within the documentation that we generate a gitea token having full `AdminScope` to manage: repository, issues, packages, etc and that a user could use it to query the API as documented here: https://docs.gitea.com/next/development/api-usage#more-on-the-authorization-header ?

pkg/util/util.go
```diff
@@ -133,3 +137,14 @@ func IsYamlFile(input string) bool {
...
+
+func GetHttpClient() *http.Client {
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		DialContext: (&net.Dialer{
+			Timeout:   5 * time.Second,
+			KeepAlive: 30 * time.Second, // from http.DefaultTransport
+		}).DialContext,
+	}
+	return &http.Client{Transport: tr, Timeout: 30 * time.Second}
```
blakeromano, line 149: This timeout seems really long but maybe I am misunderstanding

blakeromano (approved):
Looks fine as is. Just one small comment

Issue #395, opened 2024-10-04
[Suggestion]: Improve the format of the messages logged


cmoulliard:
### Have you searched for this feature request?

- [X] I searched but did not find similar requests

## Suggestion

Improve the format of the messages logged which are really long to make them shorter as they : 
- Contain many words. Idea: Add a CR after x chars, etc
- Include many repetitions:
```
Example:
Oct  4 10:49:04 INFO Checking if we should shutdown controller=localbuild controllerGroup=idpbuilder.cnoe.io controllerKind=Localbuild name=localdev name=localdev reconcileID=d1f7bb40-15fd-496a-9424-c6ebc8319bcc
Oct  4 10:49:04 INFO installing bootstrap apps to ArgoCD controller=localbuild controllerGroup=idpbuilder.cnoe.io controllerKind=Localbuild name=localdev name=localdev reconcileID=d1f7bb40-15fd-496a-9424-c6ebc8319bcc
```

![Screenshot 2024-10-03 at 18 10 27](https://github.com/user-attachments/assets/f773dfac-36f7-4f32-a671-5bd0ebf96a2b)

**Note**: In java, we log the messages as such and that improves the readability
```
__  ____  __  _____   ___  __ ____  ______ 
 --/ __ \/ / / / _ | / _ \/ //_/ / / / __/ 
 -/ /_/ / /_/ / __ |/ , _/ ,< / /_/ /\ \   
--\___\_\____/_/ |_/_/|_/_/|_|\____/___/   
2024-10-04 10:55:19,450 INFO  [io.quarkus] (main) code-with-quarkus 1.0.0-SNAPSHOT on JVM (powered by Quarkus 999-SNAPSHOT) started in 0.477s. Listening on: http://0.0.0.0:8080
2024-10-04 10:55:19,454 INFO  [io.quarkus] (main) Profile prod activated. 
2024-10-04 10:55:19,454 INFO  [io.quarkus] (main) Installed features: [argocd, cdi, kubernetes, rest, smallrye-context-propagation, vertx]
...
```

### Possible Solution

n/a

nabuskey:
Agreed. I'd like to improve the readability more. Implementation wise, it shouldn't be too difficult because we already have custom logger implemented. Just need format changes.

cmoulliard:
I also notice that some messages are not logged as (I think so) still use Logger.V (see tls.go)

PR #397, opened 2024-10-04
Change setupLog.V(1) to setupLog.info to log some missing messages

cmoulliard:
- Change setupLog.V(1) to setupLog.info to log some missing messages
- To contribute to this ticket: #395

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -134,13 +134,13 @@ func (b *Build) Run(ctx context.Context, recreateCluster bool) error {
 		return err
 	}
 
-	setupLog.V(1).Info("Getting Kube config")
+	setupLog.Info("Getting Kube config")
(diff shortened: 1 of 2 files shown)


nabuskey (commented):
I actually made these debug messages because they aren't really relevant for end users. If more verbose log messages are required, you can always increase the verbosity with the `-l` flag.

cmoulliard:
> I actually made these debug messages because they aren't really relevant for end users. If more verbose log messages are required, you can always increase the verbosity with the `-l` flag.

This is why you are using `setupLog.V(1)` ? Why don't we then use then (like for java language) a debug mode and that we log such messages using `setupLog.debug()` ?

cmoulliard:
Ok. I will close this PR as such a change is not needed

Issue #385, opened 2024-09-10
Document how to provide credentials for private Git repos


OneCricketeer:
### What is your environment, configuration, and command?

Latest version as of this issue

Running `idpbuilder create -p  https://github.com/cnoe-io/<a private repo>//ref-implementation`

### What did you do and What did you see instead?

Services started, but it was unable to clone them into Gitea showing repeated errors of `cloning repo: authentication required`

### Additional Information. Logs.

Please create documentation on how to handle creation with private stacks.

OneCricketeer:
Probably related to #237

nabuskey:
https://cloud-native.slack.com/archives/C05TN9WFN5S/p1724892781490279?thread_ts=1724877726.695989&cid=C05TN9WFN5S

I helped Abby with this. Please see if this works for you. 
I will have to document this part as well.

OneCricketeer:
@nabuskey This thread seems to be about Gitea, when I want to use a private repo from Github with the `--packages` arg.

The current workaround is to clone locally, then provide a local folder as the package

nabuskey:
We have a draft PR for supporting private repositories with SSH (private key). https://github.com/cnoe-io/idpbuilder/pull/405

Also documented here: https://cnoe.io/docs/reference-implementation/installations/idpbuilder/usage#example-commands

We need to decide what we want to do in the PR.

PR #405, opened 2024-10-11
initial ssh clone support

nabuskey:
Related to: https://github.com/cnoe-io/idpbuilder/issues/385

This adds support for cloning private repositories by SSH.

Currently, when users want to use a private repository, they need to first clone the private repository, then run idpbuilder against the local copy. This allows them to specify a private repository with ssh clone format: 

```bash
idpbuilder  create --recreate  --package git@github.com:nabuskey/stacks-private//basic/package2
```

Currently only SSH Key authentication without password is supported.

I am not sure about introducing something like this because:
1. This adds complexity and cannot possibly account for different SSH configurations. e.g. trying all valid ssh keys, accounting for passwords, etc.
2. There's already a valid workaround. Cloning a private repo with way more reliable `git clone` command.
3. In the future, it may introduce support for http auth which becomes even more complex.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/go-logr/logr v1.4.1
 	github.com/google/go-cmp v0.6.0
 	github.com/google/go-github/v61 v61.0.0
+	github.com/kevinburke/ssh_config v1.2.1-0.20231022042432-1d09c0b50564
 	github.com/spf13/cobra v1.8.0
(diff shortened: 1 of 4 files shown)


punkwalker:
Can we close this? as we have SSH clone support now?

nabuskey:
We don't have ssh clone support yet but let's close it.

Issue #410, opened 2024-10-15
[Bug]: Kind cannot pull container images from gitea when using path based routing


nabuskey:
### What is your environment, configuration, and command?

when the ` --use-path-routing` flag is used, the kind config needs to match like so:

```yaml
containerdConfigPatches:
- |-
  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."cnoe.localtest.me:8443"]
    endpoint = ["https://cnoe.localtest.me"]
  [plugins."io.containerd.grpc.v1.cri".registry.configs."cnoe.localtest.me".tls]
    insecure_skip_verify = true
```

Otherwise, the kind cluster cannot pull images references by pods.

### What did you do and What did you see instead?

Could not pull images

mouuii:
can you  add some ref to explain this problem ?

Issue #344, opened 2024-07-30
Add gitea registry tests


nabuskey:
We need e2e tests for gitea registry. 

Something like:

```
$ docker login gitea.127.0.0.1.nip.io:8443
Username: giteaadmin
Password:
WARNING! Your password will be stored unencrypted in /home/ubuntu/.docker/config.json.
Configure a credential helper to remove this warning. See
https://docs.docker.com/engine/reference/commandline/login/#credentials-store

Login Succeeded

$ docker tag docker.io/library/ubuntu:24.04 gitea.127.0.0.1:8443/giteaadmin/ubuntu:24.04

docker push gitea.127.0.0.1:8443/giteaadmin/ubuntu:24.04
The push refers to repository [gitea.127.0.0.1:8443/giteaadmin/ubuntu]
Get "https://gitea.127.0.0.1:8443/v2/": dial tcp: lookup gitea.127.0.0.1 on 127.0.0.53:53: no such host
```

cmoulliard:
As reported, we have an issue when we try to access `gitea.cnoe.localtest.me:8443` in a pod running on idpbuilder cluster gitea (ans also argocd, etc packages) as documented with this testing project: https://github.com/ch007m/idpbuilder-gitea-issue?tab=readme-ov-file

So we should add tests able to verify/validate that BUT also propose a way to also expose the services: argocd, gitea, etc using the VM/Laptop IP adddress and not the IP: `127.0.0.1` or hostname: `localhost` to avoid such an error when we call gitea within a pod

```
[curl-gitea] * IPv6: (none)
[curl-gitea] * IPv4: 127.0.0.1
[curl-gitea] *   Trying 127.0.0.1:8443...
[curl-gitea] * connect to 127.0.0.1 port 8443 from 127.0.0.1 port 60442 failed: Connection refused
[curl-gitea] * Failed to connect to gitea.cnoe.localtest.me port 8443 after 0 ms: Could not connect to server
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[curl-gitea] * closing connection #0
[curl-gitea] curl: (7) Failed to connect to gitea.cnoe.localtest.me port 8443 after 0 ms: Could not connect to server
```

dvgitit:
Is this correct, even for path based routing the docker command uses subdomain like notation for the commands?

PR #411, opened 2024-10-15
fix gitea registry for path based routing

nabuskey:
fixes: #410 
fixes: #344

### DIFF
--- a/pkg/kind/cluster_test.go
+++ b/pkg/kind/cluster_test.go
@@ -18,20 +18,20 @@ import (
 )
 
 func TestGetConfig(t *testing.T) {
-	cluster, err := NewCluster("testcase", "v1.26.3", "", "", "", util.CorePackageTemplateConfig{
-		Host: "cnoe.localtest.me",
-		Port: "8443",
-	})
-	if err != nil {
-		t.Fatalf("Initializing cluster resource: %v", err)
+
+	type tc struct {
+		host           string
+		port           string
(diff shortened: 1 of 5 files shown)


cmoulliard (commented):
The name of the parameter that we are using here: `use-path-routing` is not meaningful as in fact it allows the user to configure the internal container registry using either the FQDN of gitea or the domain name.

I suggest to rename it to something like `no-fqdn`

By the way, the documentation should be reviewed to detail what `idp` do when this parameter/Argument is selected !!

nabuskey:
> The name of the parameter that we are using here: `use-path-routing` is not meaningful as in fact it allows the user to configure the internal container registry using either the FQDN of gitea or the domain name.
> 
> I suggest to rename it to something like `no-fqdn`
> 
> By the way, the documentation should be reviewed to detail what `idp` do when this parameter/Argument is selected !!

We can revisit the flag name in the future for sure. We do need to illustrate what the flag does too.

blakeromano (approved):
I do think before we release we need some protection on if cluster was created with path based routing and I create with host based routing on the second run that we should log an error or some sort of indication that hey we are doing something wrong or this requires to be recreated... 

Talking with Manabu on slack I think we are in agreement that as long as both this and that are in the same release I am fine with approving this

nabuskey:
Opened https://github.com/cnoe-io/idpbuilder/issues/416 to track what Blake said.

Issue #386, opened 2024-09-10
[Bug]: Codespaces - failed to ensure docker network


OneCricketeer:
### What is your environment, configuration, and command?

`idpbuilder 0.7.0 go1.21.3`

Fresh installation following Quickstart Guide, trying to use Codespaces. 

### What did you do and What did you see instead?

```idpbuilder create``` fails to create a Docker network in Codespaces

```
Sep 10 01:41:21 INFO Creating kind cluster logger=setup 
Sep 10 01:41:21 INFO Runtime detected logger=setup provider=docker 
########################### Our kind config ############################
# Kind kubernetes release images https://github.com/kubernetes-sigs/kind/releases
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
  image: "kindest/node:v1.30.0"
  labels:
    ingress-ready: "true"
  extraPortMappings:
  - containerPort: 443
    hostPort: 8443
    protocol: TCP
  
containerdConfigPatches:
- |-
  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."gitea.cnoe.localtest.me:8443"]
    endpoint = ["https://gitea.cnoe.localtest.me"]
  [plugins."io.containerd.grpc.v1.cri".registry.configs."gitea.cnoe.localtest.me".tls]
    insecure_skip_verify = true
#########################   config end    ############################
Sep 10 01:41:21 INFO Creating kind cluster logger=setup cluster=localdev 
Sep 10 01:41:21 ERROR Error starting kind cluster logger=setup err=failed to ensure docker network: command "docker network create -d=bridge -o com.docker.network.bridge.enable_ip_masquerade=true -o com.docker.network.driver.mtu=1500 --ipv6 --subnet fc00:f853:ccd:e793::/64 kind" failed with error: exit status 1 
Error: failed to ensure docker network: command "docker network create -d=bridge -o com.docker.network.bridge.enable_ip_masquerade=true -o com.docker.network.driver.mtu=1500 --ipv6 --subnet fc00:f853:ccd:e793::/64 kind" failed with error: exit status 1
```

```
failed to ensure docker network: command "docker network create -d=bridge -o com.docker.network.bridge.enable_ip_masquerade=true -o com.docker.network.driver.mtu=1500 --ipv6 --subnet fc00:f853:ccd:e793::/64 kind" failed with error: exit status 1
```

Running that command directly shows an IPTables error

```
[...]

Issue #412, opened 2024-10-16
[Bug]: codespace fail to run idpbuilder create 


mouuii:
### What is your environment, configuration, and command?

env : codespace 
command :  idpbuilder create --protocol http  \
 --host ${CODESPACE_NAME}-8080.${GITHUB_CODESPACES_PORT_FORWARDING_DOMAIN} \
 --port 8080 --use-path-routing

### What did you do and What did you see instead?

err message

### Additional Information. Logs.

failed to ensure docker network: command "docker network create -d=bridge -o com.docker.network.bridge.enable_ip_masquerade=true -o com.docker.network.driver.mtu=1500 --ipv6 --subnet fc00:f853:ccd:e793::/64 kind" failed with error: exit status 1

mouuii:
@csantanapr

nabuskey:
https://github.com/kubernetes-sigs/kind/issues/3748

squidboylan:
I think this can be closed

PR #419, opened 2024-10-17
Workaround for codespaces

nabuskey:
Docker was updated in Codespaces and it caused an issue with Kind. This is a workaround for it. It just pre-creates the kind network.

For a more permanent fix, we need to update Kind once a new version with the fix is released.

https://github.com/cnoe-io/idpbuilder/issues/412
fixes: https://github.com/cnoe-io/idpbuilder/issues/386

### DIFF
--- a/.devcontainer/postCreateCommand.sh
+++ b/.devcontainer/postCreateCommand.sh
@@ -20,3 +20,4 @@ mkdir $HOME/.kube
 echo "source <(kubectl completion bash)" >> $HOME/.bashrc
 echo "alias k=kubectl" >> $HOME/.bashrc
 echo "complete -F __start_kubectl k" >> $HOME/.bashrc
+docker network create -d=bridge -o com.docker.network.bridge.enable_ip_masquerade=true -o com.docker.network.driver.mtu=1500 --subnet fc00:f853:ccd:e793::/64 kind


.devcontainer/postCreateCommand.sh
```diff
@@ -20,3 +20,4 @@ mkdir $HOME/.kube
 echo "source <(kubectl completion bash)" >> $HOME/.bashrc
 echo "alias k=kubectl" >> $HOME/.bashrc
 echo "complete -F __start_kubectl k" >> $HOME/.bashrc
+docker network create -d=bridge -o com.docker.network.bridge.enable_ip_masquerade=true -o com.docker.network.driver.mtu=1500 --subnet fc00:f853:ccd:e793::/64 kind
```
csantanapr, line 23: wow what the hell, I'm surprised by this.

csantanapr (approved):

.devcontainer/postCreateCommand.sh
OneCricketeer: Is this subnet something known ahead of time?

nabuskey: Yeah it's known in codespaces. Definitely not a permanent fix. We need to upgrade kind once a new one is released.

OneCricketeer: I was curious if it could be looked up by a script function?

nabuskey: Not too sure tbh. If you have another approach I'm definitely interested.

PR #431, opened 2024-11-01
add retry for getting credentials

nabuskey:
This adds retry logic to give it more time to retrieve credentials in E2E like here: https://github.com/cnoe-io/idpbuilder/actions/runs/11622415016/job/32368940613


csantanapr (approved):
I like the 5 retries

Issue #426, opened 2024-10-24
[Bug]: Deletion may not work properly when multiple providers are installed


nabuskey:
### What is your environment, configuration, and command?

In here: 

https://github.com/cnoe-io/idpbuilder/blob/1c68879ed7bef3da027da7a45f677a2395180695/pkg/cmd/delete/root.go#L35

Detect node is used to get the right provider to delete nodes, but this won't work if the cluster was created by podman in an environment where both docker and podman exist. 

### What did you do and What did you see instead?

We need to respect the `KIND_EXPERIMENTAL_PROVIDER` flag when deleting clusters

PR #436, opened 2024-11-06
detect kind provider correctly

nabuskey:
This PR implements kind CLI's logic for selecting providers. Previously it was possible to have a mismatch between what we select and actual provider selected by kind.

fixes: #426


cmoulliard:
The fact that you switched docker with podman with the switch case should avoid the issue that we have on a machine where docker & podman desktops are both installed. Until now I rename the docker binary to avoid such a problem

Issue #441, opened 2024-11-08
[Feature] Have a --dev parameter able to configure gitea and argocd with default: username/password


cmoulliard:
### Have you searched for this feature request?

- [X] I searched but did not find similar requests

## Suggestion

Have a `--dev` parameter that we could use when we create a new cluster and able to configure gitea and argocd with default: username/password

Why ? It is really a pain when we develop locally new packages and we recreate cluster several times to have to re-log manually using the UI of gitea/argocd instead of storing in the browser's password store the username/password used for `dev` purposes

### Possible Solution

Change the existing code

cmoulliard:
**Remark**: As argocd generates itself the initial password (=> store it part of the secret: argocd-initial-admin-secret )and hash it (=> hashed pwd is stored under: admin.password) - https://github.com/argoproj/argo-cd/blob/eb10b70e8a2e24d08137143cf01a8d9c26783b12/util/settings/settings.go#L2160-L2189, then we cannot use our own password. 

So a different strategy should be used in this case

nabuskey:
I think I'd rather name it more like kubectl. So something like `--insecure-static-password` or `--insecure-password`.

cmoulliard:
> So something like `--insecure-static-password` or `--insecure-password`.

`--insecure-static-password` is too long
`--insecure-password` is too restrictive as `--dev` or `--devMode` could help to enable some addrtional features able to debug, etc

PR #442, opened 2024-11-08
Create default passwords when dev mode is set. #441

cmoulliard:
- Add new parameter to enable/disable: dev mode
- Set the default password for gitea 
- #441

### DIFF
--- a/api/v1alpha1/localbuild_types.go
+++ b/api/v1alpha1/localbuild_types.go
@@ -47,13 +47,13 @@ type PackageConfigsSpec struct {
 
 // BuildCustomizationSpec fields cannot change once a cluster is created
 type BuildCustomizationSpec struct {
-	Protocol       string `json:"protocol,omitempty"`
-	Host           string `json:"host,omitempty"`
-	IngressHost    string `json:"ingressHost,omitempty"`
-	Port           string `json:"port,omitempty"`
-	UsePathRouting bool   `json:"usePathRouting,omitempty"`
-	SelfSignedCert string `json:"selfSignedCert,omitempty"`
-	DevMode        bool   `json:"devMode,omitempty"`
+	Protocol        string `json:"protocol,omitempty"`
+	Host            string `json:"host,omitempty"`
+	IngressHost     string `json:"ingressHost,omitempty"`
+	Port            string `json:"port,omitempty"`
+	UsePathRouting  bool   `json:"usePathRouting,omitempty"`
+	SelfSignedCert  string `json:"selfSignedCert,omitempty"`
+	StaticPasswords bool   `json:"staticPasswords,omitempty"`
 }
 
 type LocalbuildSpec struct {
--- a/api/v1alpha1/localbuild_types.go
+++ b/api/v1alpha1/localbuild_types.go
@@ -53,7 +53,7 @@ type BuildCustomizationSpec struct {
 	Port           string `json:"port,omitempty"`
 	UsePathRouting bool   `json:"usePathRouting,omitempty"`
 	SelfSignedCert string `json:"selfSignedCert,omitempty"`
-	StaticPassword bool   `json:"staticPasswords,omitempty"`
+	StaticPassword bool   `json:"staticPassword,omitempty"`
 }
 
 type LocalbuildSpec struct {
--- a/api/v1alpha1/localbuild_types.go
+++ b/api/v1alpha1/localbuild_types.go
@@ -53,6 +53,7 @@ type BuildCustomizationSpec struct {
 	Port           string `json:"port,omitempty"`
 	UsePathRouting bool   `json:"usePathRouting,omitempty"`
 	SelfSignedCert string `json:"selfSignedCert,omitempty"`
+	DevMode        bool   `json:"devMode,omitempty"`
 }
 
 type LocalbuildSpec struct {
--- a/api/v1alpha1/localbuild_types.go
+++ b/api/v1alpha1/localbuild_types.go
@@ -47,13 +47,13 @@ type PackageConfigsSpec struct {
 
 // BuildCustomizationSpec fields cannot change once a cluster is created
 type BuildCustomizationSpec struct {
-	Protocol        string `json:"protocol,omitempty"`
-	Host            string `json:"host,omitempty"`
-	IngressHost     string `json:"ingressHost,omitempty"`
-	Port            string `json:"port,omitempty"`
-	UsePathRouting  bool   `json:"usePathRouting,omitempty"`
-	SelfSignedCert  string `json:"selfSignedCert,omitempty"`
-	StaticPasswords bool   `json:"staticPasswords,omitempty"`
+	Protocol       string `json:"protocol,omitempty"`
+	Host           string `json:"host,omitempty"`
+	IngressHost    string `json:"ingressHost,omitempty"`
+	Port           string `json:"port,omitempty"`
+	UsePathRouting bool   `json:"usePathRouting,omitempty"`
+	SelfSignedCert string `json:"selfSignedCert,omitempty"`
+	StaticPassword bool   `json:"staticPasswords,omitempty"`
 }
 
 type LocalbuildSpec struct {
--- a/hack/argo-cd/argocd-rbac-dev.yaml
+++ b/hack/argo-cd/argocd-rbac-dev.yaml
@@ -1,0 +1,12 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    app.kubernetes.io/name: argocd-rbac-cm
+    app.kubernetes.io/part-of: argocd
+  name: argocd-rbac-cm
+  namespace: argocd
+data:
+  policy.csv: |
+    p, role:developer, applications, *, *, allow
+    g, developer, role:developer
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -28,7 +28,6 @@ var (
 
 type Build struct {
 	name                 string
-	devMode              bool
 	cfg                  v1alpha1.BuildCustomizationSpec
 	kindConfigPath       string
 	kubeConfigPath       string
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -44,7 +44,6 @@ type Build struct {
 
 type NewBuildOptions struct {
 	Name                 string
-	DevMode              bool
 	TemplateData         v1alpha1.BuildCustomizationSpec
 	KindConfigPath       string
 	KubeConfigPath       string
@@ -61,7 +60,6 @@ type NewBuildOptions struct {
 func NewBuild(opts NewBuildOptions) *Build {
 	return &Build{
 		name:                 opts.Name,
-		devMode:              opts.DevMode,
 		kindConfigPath:       opts.KindConfigPath,
 		kubeConfigPath:       opts.KubeConfigPath,
 		kubeVersion:          opts.KubeVersion,
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -279,5 +279,6 @@ func isBuildCustomizationSpecEqual(s1, s2 v1alpha1.BuildCustomizationSpec) bool 
 		s1.IngressHost == s2.IngressHost &&
 		s1.Port == s2.Port &&
 		s1.UsePathRouting == s2.UsePathRouting &&
-		s1.SelfSignedCert == s2.SelfSignedCert
+		s1.SelfSignedCert == s2.SelfSignedCert &&
+		s1.DevMode == s2.DevMode
 }
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -28,6 +28,7 @@ var (
 
 type Build struct {
 	name                 string
+	devMode              bool
 	cfg                  v1alpha1.BuildCustomizationSpec
 	kindConfigPath       string
 	kubeConfigPath       string
@@ -43,6 +44,7 @@ type Build struct {
 
 type NewBuildOptions struct {
 	Name                 string
+	DevMode              bool
 	TemplateData         v1alpha1.BuildCustomizationSpec
 	KindConfigPath       string
 	KubeConfigPath       string
@@ -59,6 +61,7 @@ type NewBuildOptions struct {
 func NewBuild(opts NewBuildOptions) *Build {
 	return &Build{
 		name:                 opts.Name,
+		devMode:              opts.DevMode,
 		kindConfigPath:       opts.KindConfigPath,
 		kubeConfigPath:       opts.KubeConfigPath,
 		kubeVersion:          opts.KubeVersion,
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -280,5 +280,5 @@ func isBuildCustomizationSpecEqual(s1, s2 v1alpha1.BuildCustomizationSpec) bool 
 		s1.Port == s2.Port &&
 		s1.UsePathRouting == s2.UsePathRouting &&
 		s1.SelfSignedCert == s2.SelfSignedCert &&
-		s1.DevMode == s2.DevMode
+		s1.StaticPasswords == s2.StaticPasswords
 }
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -280,5 +280,5 @@ func isBuildCustomizationSpecEqual(s1, s2 v1alpha1.BuildCustomizationSpec) bool 
 		s1.Port == s2.Port &&
 		s1.UsePathRouting == s2.UsePathRouting &&
 		s1.SelfSignedCert == s2.SelfSignedCert &&
-		s1.StaticPasswords == s2.StaticPasswords
+		s1.StaticPassword == s2.StaticPassword
 }
--- a/pkg/cmd/create/root.go
+++ b/pkg/cmd/create/root.go
@@ -20,6 +20,7 @@ import (
 const (
 	recreateClusterUsage   = "Delete cluster first if it already exists."
 	buildNameUsage         = "Name for build (Prefix for kind cluster name, pod names, etc)."
+	devModeUsage           = "When enabled, the platform will run the core packages with developer password."
 	kubeVersionUsage       = "Version of the kind kubernetes cluster to create."
 	extraPortsMappingUsage = "List of extra ports to expose on the docker container and kubernetes cluster as nodePort " +
 		"(e.g. \"22:32222,9090:39090,etc\")."
@@ -40,6 +41,7 @@ var (
 	// Flags
 	recreateCluster           bool
 	buildName                 string
+	devMode                   bool
 	kubeVersion               string
 	extraPortsMapping         string
 	kindConfigPath            string
@@ -67,6 +69,7 @@ func init() {
 	CreateCmd.PersistentFlags().StringVar(&buildName, "build-name", "localdev", buildNameUsage)
 	CreateCmd.PersistentFlags().MarkDeprecated("build-name", "use --name instead.")
 	CreateCmd.PersistentFlags().StringVar(&buildName, "name", "localdev", buildNameUsage)
+	CreateCmd.PersistentFlags().BoolVar(&devMode, "dev", false, devModeUsage)
 	CreateCmd.PersistentFlags().StringVar(&kubeVersion, "kube-version", "v1.30.3", kubeVersionUsage)
 	CreateCmd.PersistentFlags().StringVar(&extraPortsMapping, "extra-ports", "", extraPortsMappingUsage)
 	CreateCmd.PersistentFlags().StringVar(&kindConfigPath, "kind-config", "", kindConfigPathUsage)
@@ -132,6 +135,7 @@ func create(cmd *cobra.Command, args []string) error {
 
 	opts := build.NewBuildOptions{
 		Name:              buildName,
+		DevMode:           devMode,
 		KubeVersion:       kubeVersion,
 		KubeConfigPath:    kubeConfigPath,
 		KindConfigPath:    kindConfigPath,
--- a/pkg/cmd/create/root.go
+++ b/pkg/cmd/create/root.go
@@ -69,7 +69,7 @@ func init() {
 	CreateCmd.PersistentFlags().StringVar(&buildName, "build-name", "localdev", buildNameUsage)
 	CreateCmd.PersistentFlags().MarkDeprecated("build-name", "use --name instead.")
 	CreateCmd.PersistentFlags().StringVar(&buildName, "name", "localdev", buildNameUsage)
-	CreateCmd.PersistentFlags().BoolVar(&devMode, "dev", false, devModeUsage)
+	CreateCmd.PersistentFlags().BoolVar(&devMode, "dev-mode", false, devModeUsage)
 	CreateCmd.PersistentFlags().StringVar(&kubeVersion, "kube-version", "v1.30.3", kubeVersionUsage)
 	CreateCmd.PersistentFlags().StringVar(&extraPortsMapping, "extra-ports", "", extraPortsMappingUsage)
 	CreateCmd.PersistentFlags().StringVar(&kindConfigPath, "kind-config", "", kindConfigPathUsage)
--- a/pkg/cmd/create/root.go
+++ b/pkg/cmd/create/root.go
@@ -135,7 +135,6 @@ func create(cmd *cobra.Command, args []string) error {
 
 	opts := build.NewBuildOptions{
 		Name:              buildName,
-		DevMode:           devMode,
 		KubeVersion:       kubeVersion,
 		KubeConfigPath:    kubeConfigPath,
 		KindConfigPath:    kindConfigPath,
@@ -147,6 +146,7 @@ func create(cmd *cobra.Command, args []string) error {
 			IngressHost:    ingressHost,
 			Port:           port,
 			UsePathRouting: pathRouting,
+			DevMode:        devMode,
 		},
 
 		CustomPackageDirs:    absDirPaths,
--- a/pkg/cmd/create/root.go
+++ b/pkg/cmd/create/root.go
@@ -20,7 +20,7 @@ import (
 const (
 	recreateClusterUsage   = "Delete cluster first if it already exists."
 	buildNameUsage         = "Name for build (Prefix for kind cluster name, pod names, etc)."
-	devModeUsage           = "When enabled, the platform will run the core packages with developer password."
+	staticPasswordsUsage   = "Set a static password: developer to the applications: argocd and gitea."
 	kubeVersionUsage       = "Version of the kind kubernetes cluster to create."
 	extraPortsMappingUsage = "List of extra ports to expose on the docker container and kubernetes cluster as nodePort " +
 		"(e.g. \"22:32222,9090:39090,etc\")."
@@ -41,7 +41,7 @@ var (
 	// Flags
 	recreateCluster           bool
 	buildName                 string
-	devMode                   bool
+	staticPasswords           bool
 	kubeVersion               string
 	extraPortsMapping         string
 	kindConfigPath            string
@@ -69,7 +69,7 @@ func init() {
 	CreateCmd.PersistentFlags().StringVar(&buildName, "build-name", "localdev", buildNameUsage)
 	CreateCmd.PersistentFlags().MarkDeprecated("build-name", "use --name instead.")
 	CreateCmd.PersistentFlags().StringVar(&buildName, "name", "localdev", buildNameUsage)
-	CreateCmd.PersistentFlags().BoolVar(&devMode, "dev-mode", false, devModeUsage)
+	CreateCmd.PersistentFlags().BoolVar(&staticPasswords, "static-passwords", false, staticPasswordsUsage)
 	CreateCmd.PersistentFlags().StringVar(&kubeVersion, "kube-version", "v1.30.3", kubeVersionUsage)
 	CreateCmd.PersistentFlags().StringVar(&extraPortsMapping, "extra-ports", "", extraPortsMappingUsage)
 	CreateCmd.PersistentFlags().StringVar(&kindConfigPath, "kind-config", "", kindConfigPathUsage)
@@ -141,12 +141,12 @@ func create(cmd *cobra.Command, args []string) error {
 		ExtraPortsMapping: extraPortsMapping,
 
 		TemplateData: v1alpha1.BuildCustomizationSpec{
-			Protocol:       protocol,
-			Host:           host,
-			IngressHost:    ingressHost,
-			Port:           port,
-			UsePathRouting: pathRouting,
-			DevMode:        devMode,
+			Protocol:        protocol,
+			Host:            host,
+			IngressHost:     ingressHost,
+			Port:            port,
+			UsePathRouting:  pathRouting,
+			StaticPasswords: staticPasswords,
 		},
 
 		CustomPackageDirs:    absDirPaths,
--- a/pkg/cmd/create/root.go
+++ b/pkg/cmd/create/root.go
@@ -20,7 +20,7 @@ import (
 const (
 	recreateClusterUsage   = "Delete cluster first if it already exists."
 	buildNameUsage         = "Name for build (Prefix for kind cluster name, pod names, etc)."
-	staticPasswordsUsage   = "Set a static password: developer to the applications: argocd and gitea."
+	devModeUsage           = "Set the password \"developer\" for the admin user of the applications: argocd & gitea."
 	kubeVersionUsage       = "Version of the kind kubernetes cluster to create."
 	extraPortsMappingUsage = "List of extra ports to expose on the docker container and kubernetes cluster as nodePort " +
 		"(e.g. \"22:32222,9090:39090,etc\")."
@@ -41,7 +41,7 @@ var (
 	// Flags
 	recreateCluster           bool
 	buildName                 string
-	staticPasswords           bool
+	devMode                   bool
 	kubeVersion               string
 	extraPortsMapping         string
 	kindConfigPath            string
@@ -69,7 +69,7 @@ func init() {
 	CreateCmd.PersistentFlags().StringVar(&buildName, "build-name", "localdev", buildNameUsage)
 	CreateCmd.PersistentFlags().MarkDeprecated("build-name", "use --name instead.")
 	CreateCmd.PersistentFlags().StringVar(&buildName, "name", "localdev", buildNameUsage)
-	CreateCmd.PersistentFlags().BoolVar(&staticPasswords, "static-passwords", false, staticPasswordsUsage)
+	CreateCmd.PersistentFlags().BoolVar(&devMode, "dev-mode", false, devModeUsage)
 	CreateCmd.PersistentFlags().StringVar(&kubeVersion, "kube-version", "v1.30.3", kubeVersionUsage)
 	CreateCmd.PersistentFlags().StringVar(&extraPortsMapping, "extra-ports", "", extraPortsMappingUsage)
 	CreateCmd.PersistentFlags().StringVar(&kindConfigPath, "kind-config", "", kindConfigPathUsage)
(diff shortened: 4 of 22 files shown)


cmoulliard:
Question: Is it the best place to add the code able to generate a new password (= developer) when devMode is enabled - https://github.com/cnoe-io/idpbuilder/blob/71fefc7e458b4a41886a60cff4ae12eee60e1538/pkg/controllers/localbuild/argo.go#L53 ? @nabuskey

nabuskey:
Yeah after install() comes back with no errors.

cmoulliard:
I can log to the UI of gitea and argocd using as password "developer" but I don't see applications or gitea repositories

Screenshot 2024-11-08 at 17 21 23
Screenshot 2024-11-08 at 17 21 33

cmoulliard:
If what I did cannot work, we have 2 options:
- Create an Argocd PR to propose to use password instead of generating one usng a new parameter `--dev` or `--insecure-password or `--dev-password`
OR
- Add to our project the 2 Kube secrets (argocd-secret and argocd-initial-admin-secret) with content populated and where we set the property `adminAccount.Enabled` to false

nabuskey:
I am still not sure if the flag name is what we want to go with. 

Is proving static password the only thing needed for dev mode? Does it describe what it does? If we add more config changes for dev mode in the future, do we want to provide ways to control each change?

If the names I proposed in the issue are too long, can we get away with a short flag?

I do like the idea of having a static default passwords for sure.

cmoulliard:
> Is proving static password the only thing needed for dev mode? Does it describe what it does? If we add more config changes for dev mode in the future, do we want to provide ways to control each change?

If we continue to develop idpbuilder, having a devMode help a lot as we do for quarkus `devMode` able to launch testcontainers: backstage, DB, argocd, gitea, etc and where developers can access many information using dev UI - https://quarkus.io/guides/dev-mode-differences#dev-mode-features :-)

cmoulliard:
> If the names I proposed in the issue are too long, can we get away with a short flag?

Let's review that later when we have a PR which is working as `--dev` is short, enough to test the PR

cmoulliard:
> I do like the idea of having a static default passwords for sure.

I will make a try to see if that works

cmoulliard:
I tested the option 2 where the :
- Argocd `configmap/argocd-cm` includes the annotation: `admin.enabled: "false"`
- Argocd `secret/argocd-secret` sets the `admin.password` with the value of the password `developer` bcrypted using the argocd client command: `argocd account bcrypt --password developer`
```
apiVersion: v1
data:
  admin.password: JDJhJDEwJGxsdlRCbzdCQnlOaENuaHdJSXV3ZnVyYi51VXozeXNndXNsbzhINjZIaEhNR2QubEM3a1M2
  server.secretkey: Mk1BSUN3dGN1NVExUkZvZTV2Ump4QTEweTZZUkJ2RGdtMVI4UlI0OHZJaz0=
```
- Argocd `secret/argocd-initial-admin-secret` sets the `password` withe valueof the string `developer` base64 
```
data:
  password: ZGV2ZWxvcGVy
kind: Secret
  name: argocd-initial-admin-secret
  namespace: argocd
type: Opaque
```

BUT when I access the UI we got as error `login is disabled` 

Screenshot 2024-11-12 at 08 49 12

@nabuskey

cmoulliard:
> Apparently, we should be (to be investigated of course) to set the accounts directly within the argocd secret as discussed here: [argoproj/argo-cd#4096 (comment)](https://github.com/argoproj/argo-cd/issues/4096#issuecomment-751999047)

I was able to add a user `developer` with its password and to define RBAC. Still need now to add it to this PR

cmoulliard:
We cannot log using `developer`'s user and password`developer` as the argocd server reports as error 
```
time="2024-11-12T10:09:32Z" level=info msg="finished unary call with code Unauthenticated" 
error="rpc error: code = Unauthenticated 
desc = Invalid username or password" 
grpc.code=Unauthenticated grpc.method=Create grpc.service=session.SessionService grpc.start_time="2024-11-12T10:09:32Z" grpc.time_ms=530.613 span.kind=server system=grpc
```

cmoulliard:
The PR supports now to log on to the UI of Argocd using (developer/developer) with RBAC `Applications - Admin` or gitea (giteaadmin/developer)

Can you please review it ? 
@nabuskey

cmoulliard:
Apparently we cannot use for gitea `developer/developer` as we got such an error reported when we create a cluster
```
Nov 12 11:52:39 ERROR Reconciler error controller=gitrepository controllerGroup=idpbuilder.cnoe.io controllerKind=GitRepository name=argocd namespace=idpbuilder-dabou namespace=idpbuilder-dabou name=argocd reconcileID=1f80e800-276a-4b74-a868-b54ff4a26de5 err=creating repository: failed to create git repository: The repository with the same name already exists. 
Nov 12 11:52:39 ERROR Reconciler error controller=gitrepository controllerGroup=idpbuilder.cnoe.io controllerKind=GitRepository name=nginx namespace=idpbuilder-dabou namespace=idpbuilder-dabou name=nginx reconcileID=798effc7-600a-4035-8e39-eb8bcdcb79b3 err=creating repository: failed to create git repository: The repository with the same name already exists. 
Nov 12 11:52:39 ERROR Reconciler error controller=gitrepository controllerGroup=idpbuilder.cnoe.io controllerKind=GitRepository name=gitea namespace=idpbuilder-dabou namespace=idpbuilder-dabou name=gitea reconcileID=bbb8fbdd-2803-4904-b0c5-c680e2a9c0b9 err=creating repository: failed to create git repository: The repository with the same name already exists. 
```
I will then revert to the user `giteaAdmin` in the meantime

hack/argo-cd/argocd-rbac-dev.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: ConfigMap
+metadata:
+  labels:
+    app.kubernetes.io/name: argocd-rbac-cm
+    app.kubernetes.io/part-of: argocd
+  name: argocd-rbac-cm
+  namespace: argocd
+data:
+  policy.csv: |
+    p, role:developer, applications, *, *, allow
```
nabuskey, line 11: What's the purpose of having a separate account that has a very similar permissions as admin?

pkg/controllers/localbuild/argo.go
```diff
@@ -53,7 +68,74 @@ func (r *LocalbuildReconciler) ReconcileArgo(ctx context.Context, req ctrl.Reque
...
+				"accounts.developer.passwordMtime": time.Now().Format(time.RFC3339),
+			},
+		}
+		// Convert patch data to JSON
+		patchBytes, err := json.Marshal(patchData)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("Error marshalling patch data: %w", err)
+		}
+
+		kubeClient, err := k8s.GetKubeClient()
```
nabuskey, line 95: The reconciler has client already. You can just use that instead. e.g. `r.Client.`. You can also change passwords through ArgoCD api.

pkg/controllers/localbuild/argo.go
```diff
@@ -53,7 +68,74 @@ func (r *LocalbuildReconciler) ReconcileArgo(ctx context.Context, req ctrl.Reque
...
+		err = kubeClient.Get(ctx, client.ObjectKey{Name: argocdAdminSecretName, Namespace: argocdNamespace}, &s)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("getting argocd secret: %w", err)
+		}
+
+		// Patching the argocd-secret with the user's hashed password
+		err = kubeClient.Patch(ctx, &s, client.RawPatch(types.StrategicMergePatchType, patchBytes))
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("Error patching the Secret: %w", err)
+		}
```
nabuskey, line 111: You can use server side apply instead here to simplify this process.

pkg/controllers/localbuild/argo.go
```diff
@@ -15,6 +22,14 @@ import (
 //go:embed resources/argo/*
 var installArgoFS embed.FS
 
+const (
+	argocdDevModePassword         = "developer"
+	argocdAdminSecretName         = "argocd-secret"
+	argocdInitialAdminSecretName  = "argocd-initial-admin-secret"
+	argocdInitialAdminPasswordKey = "argocd-initial-admin-secret"
+	argocdNamespace               = "argocd"
```
nabuskey, line 30: These are already defined in other packages I think. Either export them or move them to globals or APIs

pkg/cmd/create/root.go
```diff
@@ -20,6 +20,7 @@ import (
 const (
 	recreateClusterUsage   = "Delete cluster first if it already exists."
 	buildNameUsage         = "Name for build (Prefix for kind cluster name, pod names, etc)."
+	devModeUsage           = "When enabled, the platform will run the core packages with developer password."
```
nabuskey, line 23: 1. I am just not sold on the name `dev mode`. idpbuilder is primary used for development and testing purposes already. I think we should use more explicit name with a short flag.
2. Even if we accept dev mode, we shouldn't limit the flag name to just passwords. dev mode could mean a lot of different things. This should be a meta flag.

pkg/k8s/util.go
```diff
@@ -2,13 +2,22 @@ package k8s
...
+	"path/filepath"
+	ctrl "sigs.k8s.io/controller-runtime"
 
 	"github.com/cnoe-io/idpbuilder/pkg/util"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+var (
+	setupLog = ctrl.Log.WithName("k8s")
```
nabuskey, line 18: I would rather not produce log messages in utility functions. Let's wrap the error instead. e.g. `fmt.Errorf("Error creating kubernetes client: %w"`

nabuskey (commented):
If you are having problems setting passwords on install, we can change them through APIs after install.

hack/argo-cd/argocd-rbac-dev.yaml
cmoulliard: This is an account for the developers and it only allows to handle applications

pkg/controllers/localbuild/argo.go
cmoulliard: > You can also change passwords through ArgoCD api.

That will require to be done in a 2nd step while here the idea is to have the account developer/developer available when packages are installed too

pkg/controllers/localbuild/argo.go
cmoulliard: Ok. I will review

pkg/cmd/create/root.go
cmoulliard: > 1. idpbuilder is primary used for development and testing purposes already. I think we should use more explicit name with a short flag.

If this is the case, then we should install gitea and argocd using a non generated password ;-)

cmoulliard: > 2\. Even if we accept dev mode, we shouldn't limit the flag name to just passwords. dev mode could mean a lot of different things. This should be a meta flag.

What about using the the parameter: `--devPwd` or `--devPassword` as boolean @nabuskey

pkg/k8s/util.go
cmoulliard: ok

cmoulliard: Fixed with: 8158551

pkg/controllers/localbuild/argo.go
cmoulliard: I removed 2 non used constants and others are non declared in another go file or package. See: 78e1e40

pkg/controllers/localbuild/argo.go
cmoulliard: What do you mean by "server side" ? My code is similar to what we do here with the Gitea - https://github.com/cnoe-io/idpbuilder/blob/f31a08e5d8f023afaeca08a9db8e88884af755e1/pkg/controllers/localbuild/gitea.go#L189 ...

nabuskey: The way it's one in Gitea is server side apply. If you look further up, you'll see it's working with unstructured object to avoid having ownership of fields we do not care about. Sometimes we see errors like "object has been updated and version doesn't match". Server side apply avoids that. 

This talks about it: https://eng.d2iq.com/blog/conflict-resolution-kubernetes-server-side-apply/

Here's where unstructured object is used:
https://github.com/cnoe-io/idpbuilder/blob/f31a08e5d8f023afaeca08a9db8e88884af755e1/pkg/controllers/localbuild/gitea.go#L163-L166

Here's another example: 
https://github.com/cnoe-io/idpbuilder/blob/d14289a40c6fed3909e1686e48d40acef4df8b8e/pkg/util/util.go#L78-L86

pkg/cmd/create/root.go
nabuskey: I am ok with `dev-pwd` or `dev-password` since we use snake case for our flags.

pkg/controllers/localbuild/argo.go
nabuskey: > > You can also change passwords through ArgoCD api.
> 
> That will require to be done in a 2nd step while here the idea is to have the account developer/developer available when packages are installed too

Yeah I am starting to think we should set them through API for both Gitea and ArgoCD in every invocation of idpbuilder. Instead of relying on manifests.

hack/argo-cd/argocd-rbac-dev.yaml
nabuskey: Can you tell me if this is inline with what you are thinking?

1. Use a random password for the developer and admin account if the dev password flag is unset.
2. Use a static, known password for the developer and admin account if the dev password flag is set.

cmoulliard: Option 2 => Use a known password for the developer and admin account if the dev password flag is set

pkg/cmd/create/root.go
cmoulliard: Finally we aligned our paths => I will then use `--dev-mode`

hack/argo-cd/argocd-rbac-dev.yaml
nabuskey: Ok that sounds good to me. Looks like Gitea static password isn't working for some reason?

cmoulliard: > Ok that sounds good to me. Looks like Gitea static password isn't working for some reason?

For gitea when using dev-mode, we should still use as user: giteaAdmin and password = developer

nabuskey: Since you are adding a user called developer to argocd, we may as well add the developer user in Gitea.

pkg/controllers/localbuild/argo.go
cmoulliard: Fixed. See: 80a785b

cmoulliard:
I pushed a commit: 3c3511d to show the `developer` username/password @nabuskey 

**Remark**: This code should be reviewed and improved as the secret containing the developer username/password is argocd-secret where the password has been bcrypted and by consequence we cannot get and decode it from the secret
This is why it has been included as such and will be displayed every time no matter if --dev-mode has been used or not

pkg/cmd/get/secrets.go
```diff
@@ -206,6 +207,18 @@ func secretToTemplateData(s v1.Secret) TemplateData {
...
+	// where the password has been bcrypted and by consequence we cannot get and decode it from the secret
+	// This is why we are going to add it here BUT it will be displayed every time no matter if --dev-mode has been used or not
+	if strings.Contains(s.Name, "gitea") {
+		data.Data["username-developer"] = "giteAdmin"
+		data.Data["password-developer"] = "developer"
+	} else if strings.Contains(s.Name, "argocd") {
+		data.Data["username-developer"] = "developer"
+		data.Data["password-developer"] = "developer"
+	}
+
```
nabuskey, line 221: We can get the localbuild object from K8s since it's essentially our way of storing configurations. You added the `DevMode` field so you can just check the value. We should change that field name btw. `devPassword` makes sense.

nabuskey:
The more I think about it, I think it's cleaner to change passwords and add users post install. This enables flows like:

1. Create a cluster without specifying the static password flag. e.g. `idpbuilder create`
2. Run the same command except we specify the static password flag. `idpbuilder create --dev-password`

cmoulliard:
> The more I think about it, I think it's cleaner to change passwords and add users post install

I'm also in favor to go down that road as I mentioned also in a different comments where we could use job or something else able to perform additional config steps or let's say post install steps.

2 remarks: 
- Such approach will nevertheless don't allow the user to watch directly using developr password the information avaialble from the web application using UI except if they use the generated password
- We should avoid to request to the user to re-apply the command `idpbuilder create` with additional flags as that will create confusion. 
A better approach would be to have a different command to install the packages and their config, execute some "actions" on a core package or package using a job, etc.
Idea about such commands could be: 
 - `idp package add <package_name> or <package1_name,package2_name2,etc>`
 - `idp package execute <job1>` // for the post-installation tasks:

pkg/cmd/get/secrets.go
cmoulliard: > We can get the localbuild object from K8s since it's essentially our way of storing configurations.

Yes but the flag `--dev-mode` or `--dev-password` is not passed at all to the command: `idp get secrets` and users will not like to perform such a command `idp get secrets --dev-mode`.

cmoulliard:
As the result about what I did and also what I discovered  imply that finally we hack the existing code to support to generate a `developer` password, should we close it and come with a better solution as proposed here: https://github.com/cnoe-io/idpbuilder/pull/442#issuecomment-2494303410

WDYT ? @nabuskey

nabuskey:
>2 remarks:
Such approach will nevertheless don't allow the user to watch directly using developr password the information avaialble from the web application using UI except if they use the generated password
    We should avoid to request to the user to re-apply the command idpbuilder create with additional flags as that will create confusion.
    A better approach would be to have a different command to install the packages and their config, execute some "actions" on a core package or package using a job, etc.
    Idea about such commands could be:
    idp package add <package_name> or <package1_name,package2_name2,etc>
    idp package execute <job1> // for the post-installation tasks:


What I meant was to do this in post-install (e.g. the core packages are ready) during the reconciliation loop.  Essentially around here: https://github.com/cnoe-io/idpbuilder/blob/03559e6871b90b834cc8d6934de74ea6f82111b7/pkg/controllers/localbuild/controller.go#L81

Instead of using manifests to set initial passwords, change them to a known value if the dev password flag is set. 

> We should avoid to request to the user to re-apply the command idpbuilder create with additional flags as that will create confusion.

I agree with this. We should have a separate command to configure / reconfigure things. 

> `package execute`
This implies we need to create a configuration file for each package. I have been resisting the idea of creating yet another configuration file. Maybe we end up needing to create one because current approach of using argocd resource hooks is not too user friendly. BUT for this particular issue of setting static passwords, it's not necessary imo.

cmoulliard:
> Instead of using manifests to set initial passwords, change them to a known value if the dev password flag is set.

I dont follow you

> Instead of using manifests to set initial passwords, change them to a known value if the dev password flag is set.

I don't set initial password using manifests. Can you then elaborate a bit here ? @nabuskey

nabuskey:
So here's the flow I am thinking.

1. We do the things we already do bootstrap the cluster. e.g. stand up kind, install argocd, gitea, etc.
2. During the installation process, we do not set static passwords. We do it the way it's done as-is.
3. Once these services are up and available, we grab the initial passwords or token, then use that to change the passwords.

nabuskey:
https://cd.apps.argoproj.io/swagger-ui#operation/AccountService_UpdatePassword

It's available through the rest api. We should be able to make a request for this I think.

cmoulliard:
> It's available through the rest api. We should be able to make a request for this I think.

This is the request used by the argocd CLI but there is no command or REST API to create a new user.

nabuskey:
It is a rest API reference page. There's an instruction here too: https://argo-cd.readthedocs.io/en/latest/developer-guide/api-docs/

I tested it and it does work.

```bash
# get initial token
curl -ik https://argocd.cnoe.localtest.me:8443/api/v1/session -d $'{"username":"admin","password":"dMQW00CssVXZWz7t"}' -H "Content-Type:application/json"

# update password
curl -X PUT -ik https://argocd.cnoe.localtest.me:8443/api/v1/account/password -d $'{"currentPassword":"dMQW00CssVXZWz7t","newPassword":"password", "name":"admin"}' -H "Content-Type:application/json" -H "Authorization: Bearer $ARGOCD_TOKEN"

# verify
curl -ik https://argocd.cnoe.localtest.me:8443/api/v1/session -d $'{"username":"admin","password":"password"}' -H "Content-Type:application/json"
```

The get secrets command looks at the initial admin secret generated by ArgoCD. If we are are to change passwords, we should make our own secrets and reference that instead.

cmoulliard:
> The get secrets command looks at the initial admin secret generated by ArgoCD. If we are are to change passwords, we should make our own secrets and reference that instead.

We should then align argocd with gitea to name the secret using the same convention
```
secret name
gitea-credential
argocd-credential
```

cmoulliard:
> What I meant was to do this in post-install (e.g. the core packages are ready) during the reconciliation loop. Essentially around here:
> 
> https://github.com/cnoe-io/idpbuilder/blob/03559e6871b90b834cc8d6934de74ea6f82111b7/pkg/controllers/localbuild/controller.go#L81

Is it the right place to put such post-installation step here as the code is part of a loop which is executed several times ....

cmoulliard:
> There's a possibility that ArgoCD is not yet reachable through Ingress-nginx. In that case, we need to retry a few times.

No need to access the API using the ingress URL but instead the kubernetes Service. Nevertheless it will be needed also to test/check if the service replies

pkg/cmd/create/root.go
cmoulliard: `--dev-mode` is a pretty common parameter used by many developer's tools and CLI which could also be used to support different additional features in the future. On the opposite `--static-creds` or `--static-credentials` parameters are probably too restrictive, don't help the users as they have no idea about what `static` word means

pkg/controllers/localbuild/controller.go
```diff
@@ -89,6 +104,48 @@ func (r *LocalbuildReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		}
 	}
 
+	if r.Config.DevMode {
```
nabuskey, line 107: In general, I think the flow here should be:

If the static field is set to true:
1. Attempt authentication by using the static password.
1. If successful, there's nothing to do so we continue with the rest of the processing logic.
1. If fails to authenticate (status code 40X, I think), we assume the password has not changed.
1. Get the current password then set it to the static password.

pkg/controllers/localbuild/controller.go
```diff
@@ -581,6 +638,165 @@ func (r *LocalbuildReconciler) requestArgoCDAppSetRefresh(ctx context.Context) e
...
+	if err != nil {
+		return fmt.Errorf("Error creating HTTP request: %v\n", err), "failed"
+	}
+	req.Header.Set("Content-Type", "application/json")
+
+	// Create an HTTP client and disable TLS verification
+	client := &http.Client{}
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	transport.TLSClientConfig.InsecureSkipVerify = true
+	client.Transport = transport
```
nabuskey, line 714: We should use this instead.

https://github.com/cnoe-io/idpbuilder/blob/f76db506e379cf4f3722ccd85ec20c471d114be1/pkg/util/util.go#L143

pkg/controllers/localbuild/controller.go
```diff
@@ -581,6 +638,165 @@ func (r *LocalbuildReconciler) requestArgoCDAppSetRefresh(ctx context.Context) e
...
+		if err != nil {
+			return fmt.Errorf("Error creating HTTP request: %v\n", err), "failed"
+		}
+		req.Header.Set("Content-Type", "application/json")
+
+		// Create an HTTP client and disable TLS verification
+		client := &http.Client{}
+		transport := http.DefaultTransport.(*http.Transport).Clone()
+		transport.TLSClientConfig.InsecureSkipVerify = true
+		client.Transport = transport
```
nabuskey, line 778: This shadows. re-use the client from above.

pkg/controllers/localbuild/controller.go
```diff
@@ -581,6 +638,165 @@ func (r *LocalbuildReconciler) requestArgoCDAppSetRefresh(ctx context.Context) e
...
+	}
+
+	resp, err := client.AdminEditUser("giteaAdmin", opts)
+	if err != nil {
+		return fmt.Errorf("cannot update gitea admin user: %w", resp.StatusCode, err), "failed"
+	}
+	return nil, "succeeded"
+}
+
+func (r *LocalbuildReconciler) updateArgocdDevPassword(ctx context.Context, adminPassword string) (error, string) {
```
nabuskey, line 691: The string returned here is not really used. I'd change the signature to return error only.

pkg/controllers/localbuild/controller.go
```diff
@@ -581,6 +638,165 @@ func (r *LocalbuildReconciler) requestArgoCDAppSetRefresh(ctx context.Context) e
...
+		payloadBytes, err := json.Marshal(payload)
+		if err != nil {
+			return fmt.Errorf("Error creating JSON payload: %v\n", err), "failed"
+		}
+
+		req, err := http.NewRequest("PUT", argocdEndpoint+"/account/password", bytes.NewBuffer(payloadBytes))
+		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", argocdSession.Token))
+		req.Header.Set("Content-Type", "application/json")
+
+		resp, err := client.Do(req)
```
nabuskey, line 752: Variable names shadow. I really do not like shadowing in longer blocks.

pkg/controllers/localbuild/controller.go
```diff
@@ -581,6 +638,165 @@ func (r *LocalbuildReconciler) requestArgoCDAppSetRefresh(ctx context.Context) e
...
+		resp, err = client.Do(req)
+		if err != nil {
+			return fmt.Errorf("Error sending request: %v\n", err), "failed"
+		}
+		defer resp.Body.Close()
+
+		if resp.StatusCode == 200 {
+			// Password verification succeeded !
+			return nil, "succeeded"
+		} else {
```
nabuskey, line 790: This else is not needed.

pkg/controllers/localbuild/controller.go
```diff
@@ -581,6 +638,165 @@ func (r *LocalbuildReconciler) requestArgoCDAppSetRefresh(ctx context.Context) e
...
+			return fmt.Errorf("### New password verification failed: %s", body), "failed"
+		}
+
+	} else {
+		return fmt.Errorf("HTTP Error: %d", resp.StatusCode), "failed"
+	}
+	return nil, "failed"
```
nabuskey, line 797: What is the meaning of returning nil (indicating no error) but failed?

pkg/controllers/localbuild/gitea.go
```diff
@@ -43,7 +45,7 @@ func RawGiteaInstallResources(templateData any, config v1alpha1.PackageCustomiza
 	return k8s.BuildCustomizedManifests(config.FilePath, "resources/gitea/k8s", installGiteaFS, scheme, templateData)
 }
 
-func giteaAdminSecretObject() corev1.Secret {
+func (r *LocalbuildReconciler) GiteaAdminSecretObject() corev1.Secret {
```
nabuskey, line 48: This doesn't need to be a method. Better to move to the utils package because it's probably needed in the get command.

pkg/controllers/localbuild/gitea.go
```diff
@@ -212,7 +215,7 @@ func getGiteaToken(ctx context.Context, baseUrl, username, password string) (str
 	return token.Token, nil
 }
 
-func giteaBaseUrl(config v1alpha1.BuildCustomizationSpec) string {
+func (r *LocalbuildReconciler) GiteaBaseUrl(config v1alpha1.BuildCustomizationSpec) string {
```
nabuskey, line 218: This doesn't need to be a method and doesn't need to be exported.

nabuskey (changes requested):
Thanks for this. Looks like there are issues with build too. Please take a look.

pkg/controllers/localbuild/controller.go
```diff
@@ -89,6 +104,48 @@ func (r *LocalbuildReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		}
 	}
 
+	if r.Config.DevMode {
+		logger.Info("DevMode is enabled")
```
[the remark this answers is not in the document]
cmoulliard: Fixed: 86b26a2

api/v1alpha1/localbuild_types.go
```diff
@@ -53,6 +53,7 @@ type BuildCustomizationSpec struct {
 	Port           string `json:"port,omitempty"`
 	UsePathRouting bool   `json:"usePathRouting,omitempty"`
 	SelfSignedCert string `json:"selfSignedCert,omitempty"`
+	DevMode        bool   `json:"devMode,omitempty"`
```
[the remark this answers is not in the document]
cmoulliard: Fixed: 35fc902

pkg/build/build.go
```diff
@@ -28,6 +28,7 @@ var (
 
 type Build struct {
 	name                 string
+	devMode              bool
```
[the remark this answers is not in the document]
cmoulliard: Fixed: 35fc902

pkg/controllers/localbuild/controller.go
cmoulliard: Fixed: 700a841

pkg/controllers/localbuild/gitea.go
cmoulliard: What do you think about this f0754f7 ?

pkg/controllers/localbuild/gitea.go
cmoulliard: See: 3ab4fd3

nabuskey (changes requested):
It's fine to have the `dev-mode` and `dev-password` flags. It needs to be represented with clear intent internally in code. The flag names are user friendly but it isn't really descriptive of what it does when referring to the option internally.

When we think about the name dev-password or dev-mode, it's not clear what that means internally. Does dev-password mean we are relaxing password rules? e.g. no special char requirements. Does dev-mode mean we disable TLS?
What we want is to set a static password for all core packages. So we should make this intent clear in code.

pkg/cmd/create/root.go
```diff
@@ -67,6 +69,7 @@ func init() {
 	CreateCmd.PersistentFlags().StringVar(&buildName, "build-name", "localdev", buildNameUsage)
 	CreateCmd.PersistentFlags().MarkDeprecated("build-name", "use --name instead.")
 	CreateCmd.PersistentFlags().StringVar(&buildName, "name", "localdev", buildNameUsage)
+	CreateCmd.PersistentFlags().BoolVar(&devMode, "dev-mode", false, devModeUsage)
```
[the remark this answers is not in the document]
cmoulliard: ok but then what will be the purpose of flag `dev-mode` if now we have 2 flags: `dev-mode` and `dev-password` ? @nabuskey

pkg/controllers/gitrepository/controller.go
```diff
@@ -143,6 +143,10 @@ func (r *RepositoryReconciler) reconcileGitRepo(ctx context.Context, repo *v1alp
 		return ctrl.Result{}, fmt.Errorf("getting git provider credentials: %w", err)
 	}
 
+	if r.Config.StaticPassword {
+		creds.password = "developer"
```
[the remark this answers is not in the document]
cmoulliard: Fixed: 570ca30

pkg/controllers/localbuild/controller.go
```diff
@@ -581,6 +638,166 @@ func (r *LocalbuildReconciler) requestArgoCDAppSetRefresh(ctx context.Context) e
...
+
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			return "", fmt.Errorf("gitea admin secret not found")
+		}
+	}
+	return string(sec.Data["password"]), nil
+}
+
+func (r *LocalbuildReconciler) updateGiteaDevPassword(ctx context.Context, adminPassword string) (error, string) {
```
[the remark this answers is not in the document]
cmoulliard: Fixed: e4831cd

pkg/controllers/localbuild/controller.go
```diff
@@ -89,6 +104,48 @@ func (r *LocalbuildReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			}
+		}
+
+		// Check if the Gitea credentials secret exists
+		giteaAdminPassword, err := r.extractGiteaAdminSecret(ctx)
+		if err != nil {
+			// Gitea admin secret is not yet available ...
+			return ctrl.Result{RequeueAfter: defaultRequeueTime}, nil
+		}
+		logger.Info("Gitea admin secret found ...")
```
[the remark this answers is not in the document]
cmoulliard: Fixed: e4831cd

pkg/controllers/localbuild/controller.go
```diff
@@ -89,6 +104,48 @@ func (r *LocalbuildReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		}
 	}
 
+	if r.Config.StaticPassword {
+		logger.V(1).Info("Dev mode is enabled")
```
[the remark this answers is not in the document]
cmoulliard: Fixed: e4831cd

pkg/util/argocd.go
```diff
@@ -0,0 +1,18 @@
+package util
+
+import (
+	"fmt"
+	"github.com/cnoe-io/idpbuilder/api/v1alpha1"
+)
+
+const (
+	ArgocdDevModePassword        = "developer"
```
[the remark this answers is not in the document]
cmoulliard: Fixed

pkg/controllers/localbuild/controller.go
```diff
@@ -89,6 +104,48 @@ func (r *LocalbuildReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+
+		// Secret containing the initial argocd password exists
+		// Lets try to update the password
+		if argocdInitialAdminPassword != "" && argocdPasswordChangeStatus == "failed" {
+			err, argocdPasswordChangeStatus = r.updateArgocdDevPassword(ctx, argocdInitialAdminPassword)
+			if err != nil {
+				return ctrl.Result{}, err
+			} else {
+				logger.V(1).Info(fmt.Sprintf("Argocd admin password change %s !", argocdPasswordChangeStatus))
+			}
```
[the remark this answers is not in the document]
cmoulliard: Function `ArgocdInitialAdminSecretObject()` moved: 2fe31a3
Signature changes for both updateArgocd|GiteaDevPassword functions: e4831cd

pkg/cmd/create/root.go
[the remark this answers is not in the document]
cmoulliard: Removed `devMode` from build.go as non used: 1cd1773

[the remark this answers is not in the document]
cmoulliard: Fixed: 70e3d37 @nabuskey

nabuskey:
E2E is failing but it's unrelated to changes. We need to make issues for:

- Test coverage for changes in this PR. We need to make sure this flow (use of static passwords) is not disturbed going forward.
- Document this flag and its purposes.

nabuskey (approved):
LGTM

cmoulliard:
> * Test coverage for changes in this PR. We need to make sure this flow (use of static passwords) is not disturbed going forward.
> 
> * Document this flag and its purposes.

I agree. I will create 2 separate issues to improve the testing coverage and documentation

cmoulliard:
e2e succeeded. I propose to merge it then

PR #455, opened 2024-11-20
handle command interrupt

nabuskey:
This PR makes idpbuilder handle interrupts correctly. Currently, idpbuilder doesn't handle signals well. For example, if you send ctrl+c while it's executing, it prints out a message saying `Finished Creating IDP Successfully!`. With this PR, it correctly says that the command execution was interrupted. It still prints command usages at the end but we can address that in another PR.

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -4,8 +4,6 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"os/signal"
-	"syscall"
 	"time"
 
 	"github.com/cnoe-io/idpbuilder/api/v1alpha1"
@@ -159,8 +157,6 @@ func (b *Build) isCompatible(ctx context.Context, kubeClient client.Client) (boo
 }
 
 func (b *Build) Run(ctx context.Context, recreateCluster bool) error {
-	managerExit := make(chan error)
-
 	setupLog.Info("Creating kind cluster")
 	if err := b.ReconcileKindCluster(ctx, recreateCluster); err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


pkg/build/build.go
```diff
@@ -266,9 +268,17 @@ func (b *Build) Run(ctx context.Context, recreateCluster bool) error {
 		return fmt.Errorf("creating localbuild resource: %w", err)
 	}
 
-	err = <-managerExit
-	close(managerExit)
-	return err
+	interrupted := make(chan os.Signal, 1)
```
greghaynes, line 271: Is there a way we could move this "out" since its really a global cancellation, not simply cancelling this subroutine. I'm used to this getting set up along with other intialization, usually in the main() method. We should be able to cancel a context there which cascades down in to cancelling this context.

nabuskey: Ah good idea. Will move it up.

cmoulliard (commented):
Should we also add  `SIGINT` to tell to the program to print interrupt and then exit ?

See: https://gobyexample.com/signals

nabuskey:
> Should we also add `SIGINT` to tell to the program to print interrupt and then exit ?
> 
> See: https://gobyexample.com/signals

My understanding is `os.Interrupt` is essentially SIGINT because that's one of the few signals guaranteed to be available on all systems. So we should be good here.

pkg/build/build.go
nabuskey: @greghaynes done.

cmoulliard (approved):

PR #461, opened 2024-11-23
add labels to applications

nabuskey:
This adds `cnoe.io/package-name` and `cnoe.io/package-type` labels to applications. Then use the label to wait for core packages to be healthy.


cmoulliard:
Excellent to add such labels !

cmoulliard (approved):

Issue #463, opened 2024-12-19
[Suggestion] Enhance how we print the secret to show them using as format: Table, Yaml or JSON


cmoulliard:
## Suggestion

Enhance how we print the secret (based on the kubernetes cli-runtime libe - see: https://github.com/kubernetes/cli-runtime/blob/master/pkg/printers/tableprinter.go#L80C6-L80C21) as we did for the Clusters to show them using as format: Table, Yaml or JSON

Examples
```
// Yaml
❯ ./idpbuilder get secrets -o yaml
- name: argocd-initial-admin-secret
  namespace: argocd
  password: bQ3e6Ja6uFRTWpXs
  token: ""
  username: admin
- name: gitea-credential
  namespace: gitea
  password: Pp}a:7,9ao'^K42>,<^8o4$9eC501e4p,Spmn;*j
  token: 72c0cbd8ea8c6006a3be970293276d0fe17efcea
  username: giteaAdmin


// Json
❯ ./idpbuilder get secrets -o json
[
  {
    "name": "argocd-initial-admin-secret",
    "namespace": "argocd",
    "username": "admin",
    "password": "bQ3e6Ja6uFRTWpXs",
    "token": ""
  },
  {
    "name": "gitea-credential",
    "namespace": "gitea",
    "username": "giteaAdmin",
    "password": "Pp}a:7,9ao'^K42>,<^8o4$9eC501e4p,Spmn;*j",
    "token": "72c0cbd8ea8c6006a3be970293276d0fe17efcea"
  }
]


// Table
❯ ./idpbuilder get secrets 
NAME                          NAMESPACE   USERNAME     PASSWORD                                   TOKEN
argocd-initial-admin-secret   argocd      admin        bQ3e6Ja6uFRTWpXs                           
gitea-credential              gitea       giteaAdmin   Pp}a:7,9ao'^K42>,<^8o4$9eC501e4p,Spmn;*j   72c0cbd8ea8c6006a3be970293276d0fe17efcea

```

cmoulliard:
PR merged. So we can close it

PR #464, opened 2024-12-19
Improve the code to print the secrets using printTable

cmoulliard:
- Improve the code to print the secrets using the cli-runtime line and printTable. 
- Refactor the code to use the print functions from the util package
- See: #463


cmoulliard:
To be investigated as test is failing.
```
=== RUN   TestOutput
    secrets_test.go:222: 
        	Error Trace:	/Users/cmoullia/code/cnoe/fork-idpbuilder/pkg/cmd/get/secrets_test.go:222
        	Error:      	Expected nil, but got: &json.SyntaxError{msg:"unexpected end of JSON input", Offset:0}
        	Test:       	TestOutput
    secrets_test.go:223: 
        	Error Trace:	/Users/cmoullia/code/cnoe/fork-idpbuilder/pkg/cmd/get/secrets_test.go:223
        	Error:      	Not equal: 
        	            	expected: 4
        	            	actual  : 0
        	Test:       	TestOutput
    secrets_test.go:242: 
        	Error Trace:	/Users/cmoullia/code/cnoe/fork-idpbuilder/pkg/cmd/get/secrets_test.go:242
        	Error:      	Not equal: 
        	            	expected: 0
        	            	actual  : 2
        	Test:       	TestOutput
    secrets_test.go:243: 
        	Error Trace:	/Users/cmoullia/code/cnoe/fork-idpbuilder/pkg/cmd/get/secrets_test.go:243
        	Error:      	Not equal: 
        	            	expected: 0
        	            	actual  : 2
        	Test:       	TestOutput
```

cmoulliard:
Question: Should we for the core package secrets display the information as such ?
```
[
  {
    "name": "argocd-initial-admin-secret",
    "namespace": "argocd",
    "data": {
      "password": "cP0HzmKo8vbzpa5L",
      "username": "admin"
    }
  }
]
```
OR
```
[
  {
    "name": "argocd-initial-admin-secret",
    "namespace": "argocd",
    "username": "admin",
    "password": "cP0HzmKo8vbzpa5L",
    }
  }
]
```

cmoulliard:
> To be investigated as test is failing.

Fixed

pkg/cmd/get/root.go
```diff
@@ -22,7 +22,7 @@ func init() {
 	GetCmd.AddCommand(ClustersCmd)
 	GetCmd.AddCommand(SecretsCmd)
 	GetCmd.PersistentFlags().StringSliceVarP(&packages, "packages", "p", []string{}, "names of packages.")
-	GetCmd.PersistentFlags().StringVarP(&outputFormat, "output", "o", "", "Output format. json or yaml.")
+	GetCmd.PersistentFlags().StringVarP(&outputFormat, "output", "o", "", "Output format: table (default if not specified), json or yaml.")
```
punkwalker, line 25: let's name the default as "table" instead of "".

pkg/cmd/get/clusters.go
```diff
@@ -74,9 +74,27 @@ func list(cmd *cobra.Command, args []string) error {
...
+	switch format {
+	case "json":
+		return util.PrintDataAsJson(clusters, outWriter)
+	case "yaml":
+		return util.PrintDataAsYaml(clusters, outWriter)
+	case "":
+		return util.PrintTable(generateClusterTable(clusters), outWriter)
+	default:
+
+		return fmt.Errorf("output format %s is not supported", format)
```
punkwalker, line 97: Can we combine this function for secret and cluster both and add it in utils package to avoid duplicate code?

punkwalker:
> Question: Should we for the core package secrets display the information as such ?
> 
> ```
> [
>   {
>     "name": "argocd-initial-admin-secret",
>     "namespace": "argocd",
>     "data": {
>       "password": "cP0HzmKo8vbzpa5L",
>       "username": "admin"
>     }
>   }
> ]
> ```
> 
> OR
> 
> ```
> [
>   {
>     "name": "argocd-initial-admin-secret",
>     "namespace": "argocd",
>     "username": "admin",
>     "password": "cP0HzmKo8vbzpa5L",
>     }
>   }
> ]
> ```

I vote for 2nd option.

pkg/cmd/get/secrets.go
```diff
@@ -144,69 +151,94 @@ func printPackageSecrets(ctx context.Context, outWriter io.Writer, kubeClient cl
...
 		return fmt.Errorf("output format %s is not supported", format)
 	}
 }
 
-func secretToTemplateData(s v1.Secret) TemplateData {
-	data := TemplateData{
+func generateSecret(s v1.Secret, isCoreSecret bool) Secret {
+	secret := Secret{
 		Name:      s.Name,
 		Namespace: s.Namespace,
```
punkwalker, line 222: We should name this `populateSecrets` instead of `generateSecret`. I felt we are creating a new secret.

pkg/cmd/get/secrets.go
```diff
@@ -42,17 +37,29 @@ var SecretsCmd = &cobra.Command{
...
+)
 
 type TemplateData struct {
 	Name      string            `json:"name"`
 	Namespace string            `json:"namespace"`
 	Data      map[string]string `json:"data"`
 }
 
+type Secret struct {
+	isCore    bool
```
punkwalker, line 54: Is the use of `isCore` field just to control the printing of core package fields? Can we do that without this?

punkwalker (changes requested):
nit: I suggest that we create a separate Printer package and Printer Interface.

cmoulliard:
> I vote for 1st option. This way we can have flexibility for complex fields.

This is what I did

pkg/cmd/get/secrets.go
cmoulliard: > Is the use of `isCore` field just to control the printing of core package fields?

It is used to figure out when we have a `core package` secret containing as values: username, password and token (for gitea)

cmoulliard: > Can we do that without this?

Not really

pkg/cmd/get/root.go
cmoulliard: I agree. Fixed: 288284e

cmoulliard:
> I suggest that we create a separate Printer package



> I suggest that we create a separate Printer package

Great idea. Fixed: 771fd89

cmoulliard:
> and Printer Interface.

Why ? Can you elaborate ?

pkg/cmd/get/secrets.go
cmoulliard: Fixed: 813c0cd

pkg/cmd/get/clusters.go
cmoulliard: > Can we combine this function for secret and cluster both and add it in utils package to avoid duplicate code?

We can do that but then there is still an issue as the Table to be generated will change according to the type: Secret, Cluster, Package and then even if the code is generic, then it will be needed to have a switch or if then statement to figure out if we generate the Table for a Secret, Cluster or Package within the package printers.

**Note**: Ideally the Table to be printed should be generated within the code of the package secret, clusters, etc

Dummy example
```
func printOutput[T any](outWriter io.Writer, input []T, format string) error {
	switch format {
	case "json":
		return printer.PrintDataAsJson(input, outWriter)
	case "yaml":
		return printer.PrintDataAsYaml(input, outWriter)
	case "table":
		// If item type of []any is Secret then we call
		return printer.PrintTable(generateClusterTable(results), outWriter)
		// If item type of []any is Secret then we call
		return printer.PrintTable(generateSecretTable(results), outWriter)
	default:
		return fmt.Errorf("output format %s is not supported", format)
	}
}
```
Idea are welcome ;-) @punkwalker

nabuskey:
> Question: Should we for the core package secrets display the information as such ?
> 
> ```
> [
>   {
>     "name": "argocd-initial-admin-secret",
>     "namespace": "argocd",
>     "data": {
>       "password": "cP0HzmKo8vbzpa5L",
>       "username": "admin"
>     }
>   }
> ]
> ```
> 
> OR
> 
> ```
> [
>   {
>     "name": "argocd-initial-admin-secret",
>     "namespace": "argocd",
>     "username": "admin",
>     "password": "cP0HzmKo8vbzpa5L",
>     }
>   }
> ]
> ```

I prefer the second format but like you said, we can discuss that in another PR / issue.

punkwalker:
> > and Printer Interface.
> 
> Why ? Can you elaborate ?

We can address this later. As of now, we can merge this PR

punkwalker (approved):

cmoulliard:
Should we wait an approval from @nabuskey @blakeromano ? @punkwalker

blakeromano (approved):

PR #467, opened 2024-12-27
fix: e2e test by updating Secret output TemplateData

punkwalker:


.goreleaser.yaml
```diff
@@ -5,7 +5,7 @@ before:
     - go mod tidy
 release:
   # Mark nightly build as prerelease based on tag
-  prerelease: '{{ contains .Tag "-nightly" }}'
+  prerelease: auto
```
nabuskey, line 8: What's this for?

pkg/cmd/get/secrets.go
```diff
@@ -46,8 +47,11 @@ var (
 )
 
 type TemplateData struct {
+	IsCore    bool              `json:"isCore"`
 	Name      string            `json:"name"`
 	Namespace string            `json:"namespace"`
+	Username  string            `json:"username"`
+	Password  string            `json:"password"`
 	Data      map[string]string `json:"data"`
 }
```
nabuskey, line 56: Why are we duplicating fields from the secret struct? I thought this struct isn't used?

tests/e2e/e2e.go
```diff
@@ -275,10 +275,10 @@ func GetBasicAuth(ctx context.Context, name string) (BasicAuth, error) {
...
 
-			for i := range secs {
-				if secs[i].Name == name {
-					out.Password = secs[i].Data["password"]
-					out.Username = secs[i].Data["username"]
+			for _, sec := range secs {
+				if sec.Name == name {
+					out.Password = sec.Password
+					out.Username = sec.Username
 					break
```
nabuskey, line 282: We are probably unmarshaing to wrong type.

.goreleaser.yaml
punkwalker: Earlier expression was not evaluating correctly and nightly releases were not marked as prerelease. So this is another way of achieving the same.

pkg/cmd/get/secrets.go
punkwalker: We can use struct from entity package in that case.

tests/e2e/e2e.go
punkwalker: I think it's the right type, the sec slice is made from TemplateData struct.

punkwalker: Changed it, Now using Secret struct from entity package

.goreleaser.yaml
nabuskey: Let's do this in another PR.

punkwalker: Done

pkg/cmd/get/secrets.go
nabuskey: This struct isn't used anywhere now, correct? We should remove it.

nabuskey (approved):
LGTM. Thanks!

PR #477, opened 2025-01-23
Fix wrong import and replace entity with types from printer

cmoulliard:
- Fix wrong import and replace entity with types from printer to fix e2e tests failing


punkwalker (approved):
LGTM 🚀

Issue #227, opened 2024-04-26
[Bug]: Cannot git clone a gitea repository using ssh


cmoulliard:
### What is your environment, configuration, and command?

We cannot git clone an internal gitea repository using ssh
```
git clone git@gitea.cnoe.localtest.me:giteaAdmin/idpbuilder-localdev-my-app-manifests.git
Cloning into 'idpbuilder-localdev-my-app-manifests'...
ssh: connect to host gitea.cnoe.localtest.me port 22: Connection refused
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

### What did you do and What did you see instead?

We should be able to ssh / git clone a gitea repository as reported here "HowTo" https://github.com/cnoe-io/idpbuilder/issues/32 if we bind the ssh port of gitea with the nodePort and that we expose too this port on the kind cluster

Until now, gitea is configured with port 22/2222

![Screenshot 2024-04-26 at 17 29 19](https://github.com/cnoe-io/idpbuilder/assets/463790/63c4505b-28e1-454c-9805-e1a791a9cea3)

cmoulliard:
We should perhaps pause the inclusion of new features till we have enough "testing coverage" about the platform using Github workflows

nimakaviani:
whats the use case for which you need to clone the gitea repository? is it for core packages or custom packages?

cmoulliard:
`whats the use case for which you need to clone the gitea repository?` For every project added post creation where users would like to use argocd + gitea or even backstage + gitea etc

Bnjoroge1:
Faced this issue as well. I fixed it by just port-forwarding from the NodePort to 22, and works fine locally. Is there a better approach to fix this in idpbuilder? Happy to look into it with some pointers!

nabuskey:
I think you need to update the `extraPortMappings` field in kind config.  We probably cannot use the default port 22 though. It needs elevated permissions and may collide with ssh service already running on the host.

cmoulliard:
We can, for sure, use a different port but then it will be needed to change the URL of the repo to git clone - https://stackoverflow.com/questions/5767850/git-on-custom-ssh-port. This is not really user friendly ...

PR #478, opened 2025-01-24
Expose the gitea ssh port on 32222

squidboylan:
Sets up gitea ssh to be exposed at 32222 on the host network, container, and service layers.

It's unclear to me if we always want to expose this port. If we decide the host port we can ensure the gitea `SSH_PORT` setting matches which makes the ssh clone URL in gitea work.

fixes https://github.com/cnoe-io/idpbuilder/issues/227

### DIFF
--- a/hack/gitea/values.yaml
+++ b/hack/gitea/values.yaml
@@ -27,6 +27,8 @@ gitea:
     server:
       DOMAIN: '{{- if .UsePathRouting -}} {{ .Host }} {{- else -}} gitea.{{- .Host }} {{- end }}'
       ROOT_URL: '{{- if .UsePathRouting }} {{- .Protocol }}://{{ .Host }}:{{ .Port }}/gitea {{- else }} {{- .Protocol }}://gitea.{{ .Host }}:{{ .Port }} {{- end }}'
+      SSH_PORT: 32222
+      SSH_LISTEN_PORT: 2222
     webhook:
       ALLOWED_HOST_LIST: private
       SKIP_TLS_VERIFY: true
@@ -34,6 +36,7 @@ gitea:
 service:
   ssh:
     type: NodePort
+    port: 32222
     nodePort: 32222
     externalTrafficPolicy: Local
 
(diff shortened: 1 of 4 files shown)


squidboylan:
I created my cluster with: `./idpbuilder create --host korhal`
cloned the repo using:
```
squid@korhal:~/hacking/go/idpbuilder$ git clone ssh://git@gitea.korhal:32222/giteaAdmin/idpbuilder-localdev-argocd.git
Cloning into 'idpbuilder-localdev-argocd'...
The authenticity of host '[gitea.korhal]:32222 ([127.0.1.1]:32222)' can't be established.
RSA key fingerprint is SHA256:yE2rNfQnjblWEZkNSsX7nyJxv1e7FasJTI/M9y7CdaY.
This key is not known by any other names
Are you sure you want to continue connecting (yes/no/[fingerprint])? yes
Warning: Permanently added '[gitea.korhal]:32222' (RSA) to the list of known hosts.
remote: Enumerating objects: 10, done.
remote: Counting objects: 100% (10/10), done.
remote: Compressing objects: 100% (8/8), done.
remote: Total 10 (delta 1), reused 0 (delta 0), pack-reused 0 (from 0)
Receiving objects: 100% (10/10), 67.20 KiB | 4.20 MiB/s, done.
Resolving deltas: 100% (1/1), done.
```
and below is what the UI shows
gitea

jessesanford:
I am ok with this but @cmoulliard did raise a good point in the thread on this issue about the port not matching what the references to the git repos are. I think as long as we continue to use the https urls for pulls and we just use the ssh for special purpose or edge cases then this is fine. I don't want to see a mix of references to both though. I think when we are building automation in idpbuilder or related stacks we should stick to the https. @nabuskey what do you think?

nabuskey:
Looks good to me.

I think https clone / push is an easier way because you can get the token and the password (could be static password) with a idpbuilder command. We should support SSH too because there's no draw back as far as I can tell.

nabuskey (approved):
lgtm

Issue #454, opened 2024-11-20
Support remote kind config ref


nabuskey:
Currently the kind config supplied via the `--kind-config` flag can refer to local files only. It should be able to reference remote kind configs as well. This is useful when a stack needs a specific kind configuration. Currently users need to clone the repository to use such stacks. e.g. `--kind-config https://github.com/stacks/kind.config`

cmoulliard:
So your idea is to do this:
- if the string passed to the flag `kind-config` starts with `https://` then we parse the Url
- We download the URL under a temporary folder
- We replace the value of  the field: `NewBuildOptions.KindConfigPath` with the new local path
- We create the kind cluster using the local konfig file

Correct ? @nabuskey

nabuskey:
We probably don't need to download it to a temp folder. I think we can do all that in-memory. Otherwise, yes that's inline with what I was thinking.

cmoulliard:
> I think we can do all that in-memory.

I don't think so as kind go lib expects to read a file - see: https://github.com/kubernetes-sigs/kind/blob/main/pkg/internal/apis/config/encoding/load.go#L31-L51

PR #480, opened 2025-01-26
add support for http and https kind-configs

squidboylan:
Fixes https://github.com/cnoe-io/idpbuilder/issues/454

### DIFF
--- a/pkg/kind/cluster.go
+++ b/pkg/kind/cluster.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"github.com/cnoe-io/idpbuilder/pkg/util"
 	"github.com/cnoe-io/idpbuilder/pkg/util/files"
+	"io"
 	"io/fs"
 	"os"
 	"strconv"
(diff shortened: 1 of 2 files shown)


cmoulliard (approved):

pkg/kind/cluster.go
```diff
@@ -70,7 +72,19 @@ func (c *Cluster) getConfig() ([]byte, error) {
 	var err error
 
 	if c.kindConfigPath != "" {
-		rawConfigTempl, err = os.ReadFile(c.kindConfigPath)
+		if strings.HasPrefix(c.kindConfigPath, "https://") || strings.HasPrefix(c.kindConfigPath, "http://") {
+			resp, err := http.Get(c.kindConfigPath)
```
nabuskey, line 76: Let's add a timeout. There is a utility function to get the http client we use for reconciliation. Maybe we can use that. Note that the timeout value is configured low because we expect endpoints to be locally available.

nabuskey (changes requested):

nabuskey (approved):
LGTM. Thank you!

This file is getting a bit too much. We need to refactor it to support easier unit testing. I'll open an issue.

PR #482, opened 2025-01-28
refactor pkg/kind

squidboylan:
refactors pkg/kind/cluster.go into pkg/kind/config.go in order to simplify testing and adds relevant tests for the new utility functions. Also cleans up some unnecessary code in pkg/kind/cluster_test.go

### DIFF
--- a/pkg/kind/config.go
+++ b/pkg/kind/config.go
@@ -1,0 +1,85 @@
+package kind
+
+import (
+	"embed"
+	"fmt"
+	"io"
+	"io/fs"
+	"os"
+	"strings"
+
+	"github.com/cnoe-io/idpbuilder/api/v1alpha1"
+)
+
+type PortMapping struct {
+	HostPort      string
+	ContainerPort string
+}
+
+type TemplateConfig struct {
+	v1alpha1.BuildCustomizationSpec
+	KubernetesVersion string
+	ExtraPortsMapping []PortMapping
+	RegistryConfig    string
+}
+
+//go:embed resources/* testdata/custom-kind.yaml.tmpl
+var configFS embed.FS
+
+func loadConfig(path string, httpClient HttpClient) ([]byte, error) {
+	var rawConfigTempl []byte
+	var err error
+	if path != "" {
+		if strings.HasPrefix(path, "https://") || strings.HasPrefix(path, "http://") {
+			resp, err := httpClient.Get(path)
+			if err != nil {
+				return nil, fmt.Errorf("fetching remote kind config: %w", err)
+			}
+			defer resp.Body.Close()
+			if !(resp.StatusCode < 300 && resp.StatusCode >= 200) {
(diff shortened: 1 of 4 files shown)


squidboylan:
doing this was suggested in https://github.com/cnoe-io/idpbuilder/pull/480#pullrequestreview-2576916082

pkg/kind/config.go
```diff
@@ -0,0 +1,73 @@
...
+			if err != nil {
+				return nil, fmt.Errorf("fetching remote kind config: %w", err)
+			}
+			defer resp.Body.Close()
+			rawConfigTempl, err = io.ReadAll(resp.Body)
+			if err != nil {
+				return nil, fmt.Errorf("reading remote kind config body: %w", err)
+			}
+		} else {
+			rawConfigTempl, err = os.ReadFile(path)
```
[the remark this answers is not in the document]
squidboylan: Added these unit tests and actually found a bug with non 2XX responses when fetching a kind-config. Let me know what you think

pkg/kind/config.go
```diff
@@ -0,0 +1,90 @@
...
+	ExtraPortsMapping []PortMapping
+	RegistryConfig    string
+}
+
+//go:embed resources/* testdata/custom-kind.yaml.tmpl
+var configFS embed.FS
+
+type HttpClient interface {
+	Get(url string) (resp *http.Response, err error)
+}
```
nabuskey, line 32: I think we can move this to cluster.go and make it one of fields in the Cluster struct.

It would look something like:

```go
type Cluster struct {
	provider          IProvider
	httpClient        HttpClient
	name              string
	kubeVersion       string
	kubeConfigPath    string
	kindConfigPath    string
	extraPortsMapping string
	registryConfig    []string
	cfg               v1alpha1.BuildCustomizationSpec
}
```

In here: https://github.com/cnoe-io/idpbuilder/blob/64d93ace30b9700c81920c091ff24926066106c0/pkg/kind/cluster.go#L152-L170

We can call `util.GetHttpClient()`. When we test, use the mock client.

pkg/kind/cluster_test.go
```diff
@@ -202,69 +195,3 @@ func TestGetConfigCustom(t *testing.T) {
...
-}
-
-func (n *NodeMock) Command(command string, args ...string) exec.Cmd {
-	argsMock := append([]string{command}, args...)
-	mockArgs := n.Called(argsMock)
-	return mockArgs.Get(0).(exec.Cmd)
-}
-
-func (n *NodeMock) String() string {
-	args := n.Called()
```
nabuskey, line 248: I think we can leave mockProvider stuff  as they are. I know they are not used right now.

nabuskey (commented):
Thanks for this! Please take look at my comments. Just a small improvements we can make.

nabuskey (approved):

Issue #479, opened 2025-01-24
[Bug]: Release job fails - Private Key is in OpenSSH format, but only PKCS#8 is supported. See https://github.com/gr2m/universal-github-app-jwt#private-key-formats


cmoulliard:
## Issue

The job to release fails: https://github.com/cnoe-io/idpbuilder/actions/runs/12945012192/job/36106860618 during the step of the flow to `     `Generate a homebrew tap update token`

```
7s
Run actions/create-github-app-token@v1
owner and repositories not set, creating token for the current repository ("idpbuilder")
Failed to create token for "idpbuilder" (attempt 1): [universal-github-app-jwt] Private Key is in OpenSSH format, but only PKCS#8 is supported. See https://github.com/gr2m/universal-github-app-jwt#private-key-formats
Failed to create token for "idpbuilder" (attempt 2): [universal-github-app-jwt] Private Key is in OpenSSH format, but only PKCS#8 is supported. See https://github.com/gr2m/universal-github-app-jwt#private-key-formats
Failed to create token for "idpbuilder" (attempt 3): [universal-github-app-jwt] Private Key is in OpenSSH format, but only PKCS#8 is supported. See https://github.com/gr2m/universal-github-app-jwt#private-key-formats
Failed to create token for "idpbuilder" (attempt 4): [universal-github-app-jwt] Private Key is in OpenSSH format, but only PKCS#8 is supported. See https://github.com/gr2m/universal-github-app-jwt#private-key-formats
Error: [universal-github-app-jwt] Private Key is in OpenSSH format, but only PKCS#8 is supported. See https://github.com/gr2m/universal-github-app-jwt#private-key-formats
    at getToken (/home/runner/work/_actions/actions/create-github-app-token/v1/dist/main.cjs:38662:11)
    at githubAppJwt (/home/runner/work/_actions/actions/create-github-app-token/v1/dist/main.cjs:38705:23)
    at getAppAuthentication (/home/runner/work/_actions/actions/create-github-app-token/v1/dist/main.cjs:38868:37)
    at hook4 (/home/runner/work/_actions/actions/create-github-app-token/v1/dist/main.cjs:39166:37)
    at newApi (/home/runner/work/_actions/actions/create-github-app-token/v1/dist/main.cjs:37967:36)
    at getTokenFromRepository (/home/runner/work/_actions/actions/create-github-app-token/v1/dist/main.cjs:39491:26)
[...]

PR #484, opened 2025-01-28
fix: failing release job

punkwalker:
For homebrew-tap token create-token GH action requires repository to be specified otherwise it will create token only for the current repository. 
Ref: https://github.com/actions/create-github-app-token?tab=readme-ov-file#repositories

fixes: #479


nabuskey (approved):
lgtm

Issue #393, opened 2024-10-03
Document passing credentials to pull from private registries


nabuskey:
Steps are described here. https://github.com/cnoe-io/idpbuilder/issues/229

For ECR:

```bash
aws ecr get-login-password --region us-west-2 | docker login --username AWS --password-stdin 1111.dkr.ecr.<REGION>.amazonaws.com
idpbuilder create --recreate --kind-config kind-registry.yaml

```
kind yaml:
```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
  image: "kindest/node:v1.30.0"
  labels:
    ingress-ready: "true"
  extraPortMappings:
  - containerPort: 443
    hostPort: 8443
    protocol: TCP
  extraMounts:
  - containerPath: /var/lib/kubelet/config.json
    hostPath: /home/ubuntu/.docker/config.json

containerdConfigPatches:
- |-
  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."gitea.cnoe.localtest.me:8443"]
    endpoint = ["https://gitea.cnoe.localtest.me"]
  [plugins."io.containerd.grpc.v1.cri".registry.configs."gitea.cnoe.localtest.me".tls]
    insecure_skip_verify = true
```

PR #487, opened 2025-02-04
Add registry-config option

squidboylan:
The new registry-config option when used alone looks up the default
podman and docker registry config files and mounts them in the kind
container if they exist. Optionally a user can pass a specific file to
the flag to specify a custom path to mount as the registry config.

### DIFF
--- a/pkg/kind/cluster.go
+++ b/pkg/kind/cluster.go
@@ -39,6 +39,7 @@ type Cluster struct {
 	kubeConfigPath    string
 	kindConfigPath    string
 	extraPortsMapping string
+	registryConfig    []string
 	cfg               v1alpha1.BuildCustomizationSpec
 }
 
@@ -60,6 +61,7 @@ type TemplateConfig struct {
 	v1alpha1.BuildCustomizationSpec
 	KubernetesVersion string
 	ExtraPortsMapping []PortMapping
+	RegistryConfig    string
 }
 
 //go:embed resources/*
@@ -108,11 +110,25 @@ func (c *Cluster) getConfig() ([]byte, error) {
 		}
 	}
 
+	registryConfig := ""
+	for _, s := range c.registryConfig {
+		path := os.ExpandEnv(s)
+		if _, err := os.Stat(path); err == nil {
+			registryConfig = path
+			break
+		}
+	}
+
+	if len(c.registryConfig) > 0 && registryConfig == "" {
+		return nil, errors.New("--registry-config flag used but no registry config was found")
+	}
+
 	var retBuff []byte
 	if retBuff, err = files.ApplyTemplate(rawConfigTempl, TemplateConfig{
 		BuildCustomizationSpec: c.cfg,
 		KubernetesVersion:      c.kubeVersion,
 		ExtraPortsMapping:      portMappingPairs,
+		RegistryConfig:         registryConfig,
 	}); err != nil {
 		return nil, err
 	}
@@ -133,7 +149,7 @@ func (c *Cluster) getConfig() ([]byte, error) {
 	return retBuff, nil
 }
 
-func NewCluster(name, kubeVersion, kubeConfigPath, kindConfigPath, extraPortsMapping string, cfg v1alpha1.BuildCustomizationSpec, cliLogger logr.Logger) (*Cluster, error) {
+func NewCluster(name, kubeVersion, kubeConfigPath, kindConfigPath, extraPortsMapping string, registryConfig []string, cfg v1alpha1.BuildCustomizationSpec, cliLogger logr.Logger) (*Cluster, error) {
 	detectOpt, err := util.DetectKindNodeProvider()
 	if err != nil {
 		return nil, err
@@ -148,6 +164,7 @@ func NewCluster(name, kubeVersion, kubeConfigPath, kindConfigPath, extraPortsMap
 		kubeVersion:       kubeVersion,
 		kubeConfigPath:    kubeConfigPath,
 		extraPortsMapping: extraPortsMapping,
+		registryConfig:    registryConfig,
 		cfg:               cfg,
 	}, nil
 }
--- a/pkg/kind/cluster_test.go
+++ b/pkg/kind/cluster_test.go
@@ -21,6 +21,7 @@ func TestGetConfig(t *testing.T) {
 	type tc struct {
 		host           string
 		port           string
+		registryConfig []string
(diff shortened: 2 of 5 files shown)


squidboylan:
This is an attempt to simplify #393 , if we would prefer not to have this in the cli I can close this and document the custom kind config solution to #393

cmoulliard (approved):
I like a lot this PR as it avoids for the user to hack the kind cfg file :-)

nabuskey:
As we expose more and more kind config through CLI flags that directly corresponds to kind configurations and CLI flags, are we making it easier for people to use this tool? I think when we add flags, we should add a bit more value than exposing kind configuration options. 
Since this functionality is already available by providing a kind config file, I think we should add more value by making it easy to mount registry credentials specifically because it's a very common use case.

Thinking about how the problem of credentials mounting is solved, if we expose kind config flag directly this is what that looks like:
```
idpbuilder create --extraMountsUsage /registry/config.json:/var/lib/kubelet/config.json
```

Do we want users to enter this every time?  We can guess where registry configurations are located because docker and podman store them in  `$XDG_RUNTIME_DIR/containers/auth.json`.

To me, it makes more sense to provide a boolean flag indicating if the user wants to mount registry credentials and we do the rest.

cmoulliard:
> As we expose more and more kind config through CLI flags that directly corresponds to kind configurations and CLI flags, are we making it easier for people to use this tool? I think when we add flags, we should add a bit more value than exposing kind configuration options.

This is a fair point you reported here as yoou ask finally the question "What should be the boundary between what we propose as CLI flags/parameters vs passing such information using `kind config file` vs `something else` that users could use to override parameters. Such a something else could be: env variable, idpbuilder yaml config file, etc

Why this is not the purpose of this ticket, it is nevertheless important that we keep in mind such an aspect and that we review what we did or will in a separate ticket to edict rules about that. WDYT ? @nabuskey

cmoulliard:
> To me, it makes more sense to provide a boolean flag indicating if the user wants to mount registry credentials and we do the rest.

As the ultimate goal is to mount the credentials of the user, then having a flag pointing to the config.json or auth.json file (docker vs podman) is probably more appropriate

blakeromano:
> As we expose more and more kind config through CLI flags that directly corresponds to kind configurations and CLI flags, are we making it easier for people to use this tool? I think when we add flags, we should add a bit more value than exposing kind configuration options.


I am 100% in agreement here

squidboylan:
> > To me, it makes more sense to provide a boolean flag indicating if the user wants to mount registry credentials and we do the rest.

>  As the ultimate goal is to mount the credentials of the user, then having a flag pointing to the config.json or auth.json file (docker vs podman) is probably more appropriate

Yeah I was trying to take an approach somewhat between the two last night where we could have a `--registry-config` that optionally takes a path to a registry config but by default looks for the `$XDG_RUNTIME_DIR/containers/auth.json` file and if it doesnt find it then looks for the `~/.docker/config.json` file but cobra doesn't really provide a great way to do this. I'll put something up later today, either a hacky solution around what I'm describing or one of the two suggestions above.

squidboylan:
This is kinda hacky but seems reasonable to me. The flag can be used the following ways:
1. If it is not used no registry config is mounted
2. If it is passed on its own (`--registry-config`) it checks for the default podman and docker registry config paths and uses them if they exist (podman has precedence)
3. If it is passed with a specific file it uses that if it exists (`--registry-config="/home/foo/registry/config.json"`)

It is a little weird that this can take multiple `--registry-config` flags to specify multiple options but it doesnt seem like a deal breaker (and it makes implementation and understanding how this works a bit easier).

pkg/kind/cluster.go
```diff
@@ -108,11 +110,23 @@ func (c *Cluster) getConfig() ([]byte, error) {
 		}
 	}
 
+	var registryConfig string = func() string {
```
nabuskey, line 113: Does this need to be structured this way? I prefer a more simple way to set a variable value

pkg/kind/cluster_test.go
nabuskey, line 1: We should add  test for this.

pkg/kind/cluster.go
```diff
@@ -108,11 +110,23 @@ func (c *Cluster) getConfig() ([]byte, error) {
...
+		for _, s := range c.registryConfig {
+			path := os.ExpandEnv(s)
+			println(path)
```
nabuskey, line 116: Let's remove this. We could log as a debug message if needed.

pkg/kind/cluster.go
```diff
@@ -108,11 +110,23 @@ func (c *Cluster) getConfig() ([]byte, error) {
...
+			if _, err := os.Stat(path); err == nil {
+				return path
```
nabuskey, line 118: If an error is returned, this ignore the error correct?

nabuskey (commented):
The overall approach looks great to me. Thank you so much for this. Please take a look at my comments.

pkg/kind/cluster.go
squidboylan: Nope! Agreed, this should just be simplified, I had this pattern previously and forgot to simplify it

pkg/kind/cluster_test.go
squidboylan: Added a simple test in `TestGetConfig`

pkg/kind/cluster.go
```diff
@@ -108,11 +110,25 @@ func (c *Cluster) getConfig() ([]byte, error) {
...
+	for _, s := range c.registryConfig {
+		path := os.ExpandEnv(s)
+		if _, err := os.Stat(path); err == nil {
+			registryConfig = path
+			break
+		}
+	}
+
+	if len(c.registryConfig) > 0 && registryConfig == "" {
+		setupLog.Info("--registry-config flag used but no registry config was found")
```
nabuskey, line 123: Do you think it's a good idea to fail here? We think the user's intention was to mount registry configuration, but it does not exist. If we do not fail, the cluster is created with configuration that does not match what the user wanted it to be. So it might be a better idea to fail early and ask the user for a valid configuration.

squidboylan: Yeah I think that's a good idea, i threw a `os.Exit(1)` in that block

nabuskey: I think it's better to return an error and let the error cascade out to cobra to have an unified experience for users. Right now all fatal errors go out like that.

Issue #444, opened 2024-11-09
Document minimum requirement


nabuskey:
We need to document the minimum machine spec need to use idpbuilder with core packages only.

```
  Resource           Requests     Limits
  --------           --------     ------
  cpu                1150m (14%)  100m (1%)
  memory             508Mi (1%)   390Mi (1%)
  ephemeral-storage  0 (0%)       0 (0%)
  hugepages-1Gi      0 (0%)       0 (0%)
  hugepages-2Mi      0 (0%)       0 (0%)
```

According to this, it should be 2 vCPUs and 1 GB memory. But realistically, it should have at least 4 vCPUs and 4 GB ram available to ensure whatever users deploy actually works.

PR #493, opened 2025-03-03
Add minimum requirements and private registry docs

squidboylan:
Addresses #444 and #393

Not sure if this is what we had in mind for the docs, if these should be moved elsewhere let me know

### DIFF
--- a/docs/minimum-requirements.md
+++ b/docs/minimum-requirements.md
@@ -1,0 +1,4 @@
+# Minimum Requirements
+
+The requirements for your cluster will depend on what it is running but we
+recommend a minimum of 4 CPU cores and 4GiB of RAM.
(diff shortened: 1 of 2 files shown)


punkwalker:
As of now we can keep it in the docs folder, we will rearrange it later. Also, do you think we should add a sample registry config like shown in #393 or a link to containerd registry config docs?

squidboylan:
> Also, do you think we should add a sample registry config like shown in https://github.com/cnoe-io/idpbuilder/issues/393 or a link to containerd registry config docs?

the file shown in that PR is actually the kind config and that is what the `--registry-config` templates out for users, there shouldn't need to be any other config. The only other thing users will need to do is create the registry config by doing a `docker login` or `podman login`.

punkwalker (approved):
LGTM 🚀

PR #495, opened 2025-03-07
fix bugs around ingress, tls, and dev-password

squidboylan:
This PR fixes a few bugs around using the `--ingress-host-name` flag, `--dev-password` flag, and `--use-path-routing` flag. I initially wanted to separate these but decided they make sense to keep together since they're all a little bit interdependent.

Before this change using the `--ingress-host-name` flag would cause argo to fail to fetch repos from gitea due to missing SANs in the TLS cert. It also fixes a bug where we're hardcoding the domain for argocd and gitea to `cnoe.localtest.me` in the `--dev-password` codepath.

### DIFF
--- a/pkg/controllers/localbuild/gitea.go
+++ b/pkg/controllers/localbuild/gitea.go
@@ -157,7 +157,7 @@ func (r *LocalbuildReconciler) setGiteaToken(ctx context.Context, secret corev1.
 // gitea URL reachable within the cluster with proper coredns config. Mainly for argocd
 func giteaInternalBaseUrl(config v1alpha1.BuildCustomizationSpec) string {
 	if config.UsePathRouting {
-		return fmt.Sprintf(util.GiteaSvcURL, config.Protocol, "", config.Host, config.Port, "/gitea")
+		return fmt.Sprintf(util.GiteaURLTempl, config.Protocol, "", config.Host, config.Port, "/gitea")
 	}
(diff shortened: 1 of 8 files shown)


squidboylan:
Marking this as a draft, i think there's more bugs here and i wanna make sure I'm solving them right

pkg/controllers/localbuild/gitea.go
```diff
@@ -157,7 +157,7 @@ func (r *LocalbuildReconciler) setGiteaToken(ctx context.Context, secret corev1.
 // gitea URL reachable within the cluster with proper coredns config. Mainly for argocd
 func giteaInternalBaseUrl(config v1alpha1.BuildCustomizationSpec) string {
```
squidboylan, line 158: This could be simplified a bit to just use the pkg/util/gitea.go `GiteaBaseUrl` function, not sure if we want that change in this PR though, it's already getting a bit bloated

punkwalker: I think this function is specific for returning in-cluster URL and I can see it is only used once. And the way this PR changes it's logic, it is identical to `pkg/util/gitea.go/GiteaBaseUrl` function. So we should use that and get rid of this function.

punkwalker (commented):
The changes LGTM. I am not sure how/if these changes will impact the current assumptions that we have made in configuration of various components. Let's see how e2e test perform on this PR.

punkwalker (approved):
LGTM 🚀

PR #496, opened 2025-03-07
Restrict ingress port forwarding when using --dev-password

squidboylan:
forwarding our ingress to the host node and listening on 0.0.0.0 can potentially be dangerous, particularly if a user is using the --dev-password flag to use an insecure static password. In that case someone could potentially access their gitea/argocd install and run arbitrary pods in the cluster.


nabuskey (approved):
LGTM

Issue #362, opened 2024-08-21
Feature: add a flag for `--kubeconfig` or support `KUBECONFIG` environment variable


punkwalker:
### Have you searched for this feature request?

- [X] I searched but did not find similar requests

### Problem Statement

Currently create command uses `$HOME/.kube/config` for writing kubeconfig of Kind cluster. And get command also uses same default path for interacting with cluster.

This works fine but there are following scenarios where this behavior is not useful.
**Story 1:**
I don't want the local cluster kubeconfig in default location.

**Story 2:**
I am using context switching tools like `kubeswitch` which allows me to create sticky context for each terminal. In this case, If I create multiple builds, the `get secrets` command keeps fetching secrets from the cluster for which the context is set in the default kubeconfig.

**Story 3:**
I want to have `kubectl` like experience of using non-default `kubeconfig` when I am interacting with the cluster.

### Possible Solution

As a quick fix, the logic of `kubeConfigPath` in [create](https://github.com/cnoe-io/idpbuilder/blob/a4a4383295ee458f719f884d858030160e513bb2/pkg/cmd/create/root.go#L74) and [get secret](https://github.com/cnoe-io/idpbuilder/blob/a4a4383295ee458f719f884d858030160e513bb2/pkg/cmd/get/secrets.go#L59) command can be modified to check `KUBECONFIG` environment variable as follows:
```
	kubeConfigPath := filepath.Join(homedir.HomeDir(), ".kube", "config")
	kubeConfigPathEnv := os.Getenv("KUBECONFIG")

	if kubeConfigPathEnv != "" {
		kubeConfigPath = kubeConfigPathEnv
	}
```

Long term, we should add support for `--kubeconfig` flag and then set the value to opts. Eventually it can be used for #74

### Alternatives Considered

Currently. No alternative than using default `kubeconfig` file

PR #497, opened 2025-03-09
fix bugs around kube config path flags

squidboylan:
This centralizes the GetKubeClient and GetKubeConfig functions so that we're always using the right kube config path

### DIFF
--- a/pkg/cmd/get/clusters.go
+++ b/pkg/cmd/get/clusters.go
@@ -63,18 +63,17 @@ func populateClusterList() ([]idpTypes.Cluster, error) {
 		return nil, err
 	}
 
-	kubeConfig, err := helpers.GetKubeConfig()
+	kubeConfig, err := util.GetKubeConfig()
 	if err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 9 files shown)


punkwalker:
Thank you @squidboylan for working on this. This LGTM. Only small query: Would this fix #362 as well?

squidboylan:
> Would this fix https://github.com/cnoe-io/idpbuilder/issues/362 as well?

I dont think so, this only fixes the `--kubeconfig` flag for the `get *` commands, the `create` command still doesn't have that flag. This change does get us closer to being able to support that flag though as it should reduce some of the duplicate code that generates the kubeconfig path and builds a kube client. I can probably take a look at fixing that issue sometime soon.

punkwalker:
> > Would this fix #362 as well?
> 
> I dont think so, this only fixes the `--kubeconfig` flag for the `get *` commands, the `create` command still doesn't have that flag. This change does get us closer to being able to support that flag though as it should reduce some of the duplicate code that generates the kubeconfig path and builds a kube client. I can probably take a look at fixing that issue sometime soon.

Great. Meanwhile, we can merge this one. Thanks for confirming.

punkwalker (approved):
🚀 LGTM

PR #499, opened 2025-03-13
don't set the admin password repeatedly when using --dev-password

squidboylan:
Every time we change the admin password in argocd it clears the active user sessions. This makes iteration with developing stacks a bit tedious as you end up running `idpbuilder --dev-password -p stackname` a lot and you have to log back into argo every time. This is a minor inconvenience but I don't see any reason we can't fix it.


cmoulliard (approved):

cmoulliard:
Many thanks for the PR @squidboylan

Issue #415, opened 2024-10-16
Feature: support private deployment, without external network access.


mouuii:
### Have you searched for this feature request?

- [X] I searched but did not find similar requests

### Problem Statement

can we support Private deployment, without external network access.？

### Possible Solution

Private deployment, without external network access.

nabuskey:
Hey @mouuii,  

Yes, we received a few requests to support air gapped environments and I'd love to support that use case. Related to: https://github.com/cnoe-io/idpbuilder/issues/402

mouuii:
@nabuskey  got it

Issue #402, opened 2024-10-09
Feature: Use local images if available


nabuskey:
### Have you searched for this feature request?

- [X] I searched but did not find similar requests

### Problem Statement

When idpbuilder spins up a kind cluster, it needs to pull images from a registry at least once. At least four images must be pulled every time: argocd, redis, ingress-nginx, and gitea. It would be great if we can use the locally available images instead to save on network usages. 

### Possible Solution

You could do this with the `-c` flag but that's not really friendly.

We could use the kind's load utility to load images to the nodes. https://github.com/kubernetes-sigs/kind/blob/10e058c534cf6b8546be70f1eea2e414a08ff770/pkg/cluster/nodeutils/util.go#L80

cmoulliard:
Can a proxy registry be an alternative https://github.com/rpardini/docker-registry-proxy?tab=readme-ov-file#kind-cluster ? @nabuskey

PR #501, opened 2025-03-14
feature: attempt to import images from the host

squidboylan:
This is an attempt to address https://github.com/cnoe-io/idpbuilder/issues/402 . Throwing this up here to get thoughts on this approach, there's still more work to do including supporting package customization and tests. I considered using https://github.com/kubernetes-sigs/kind/blob/10e058c534cf6b8546be70f1eea2e414a08ff770/pkg/cluster/nodeutils/util.go#L80 like described in the issue but that seems to require a lot more code that is already implemented in kind, and this was the easiest way i could think of reusing that kind code. I admit calling the kind cobra command directly feels like a bit of a hack but at least it's likely to be stable? This also doesnt get us all the way to fixing https://github.com/cnoe-io/idpbuilder/issues/415 because it doesnt support images referenced by digest because of https://github.com/kubernetes-sigs/kind/issues/2394 .


nabuskey:
Hmmm I wonder if we should use kind load or let another cache program do this for us. @punkwalker What do you think?

cmoulliard:
> Hmmm I wonder if we should use kind load

As mentioned by Cale, `kind load` do a lot of things under the hood to upload a tarball of an image as it must access the kindnet container to execute some internal commands.

punkwalker:
> Hmmm I wonder if we should use kind load or let another cache program do this for us. @punkwalker What do you think?

I feel we should use kind for this use-case. We don't want to implement logic which can be handled by kind. I usually use a script to load the images I need when idpbuilder creates cluster.

```

#!/bin/bash

# Check if kind is installed
if ! command -v kind &> /dev/null; then
    echo "Error: kind is not installed"
    exit 1
fi

# Check if docker is installed and running
if ! docker info &> /dev/null; then
    echo "Error: Docker is not running or not installed"
    exit 1
fi

# Get the kind cluster name, default to "kind" if not specified
CLUSTER_NAME=${1:-localdev}

# Check if the cluster exists
if ! kind get clusters | grep -q "^${CLUSTER_NAME}$"; then
    echo "Error: Cluster ${CLUSTER_NAME} does not exist"
    exit 1
fi

# Function to load images
load_images() {
    local images=("$@")
    for image in "${images[@]}"; do
        echo "Loading image: $image"
        docker pull "$image"
        kind load docker-image "$image" --name "$CLUSTER_NAME"
    done
}

# List of images to load
# Add or modify images as needed
IMAGES=(
    "alpine/git:v2.45.1"
    "golang:1.22-alpine"
    "alpine:3.20"
    "alpine/k8s:1.31.0"
    "quay.io/argoproj/argoexec:v3.5.4"
    "gcr.io/kaniko-project/executor:aca3b966a3376c998f4824162912ba5ea04c9d4f"
    # Add more images here
)

# Load the images
load_images "${IMAGES[@]}"

echo "Images loaded successfully into kind cluster: $CLUSTER_NAME"
```

punkwalker:
> > @squidboylan I feel we should use kind for this use-case
> 
> @punkwalker I think there might be some confusion. In this PR I am using kind for this, however I'm not using the CLI as a user doesn't necessarily have it installed, instead I'm using the kind go package which we already depend on and I use it in a way that simulates running the kind cli how you show in your example.

Yeah, I was referring to kind CLI. 
IMO, If we need to add this, we should keep it optional with a flag instead of implicitly importing images.

Issue #144, opened 2024-02-01
A guide for running it in a remote machine


nabuskey:
Ned a guide for running it in a remote machine because Docker desktop absolutely kills my CPU for some reason. I have been running it on a remote EC2 machine with SSH port forwarding and it works perfectly. We need a guide on this so that others with less powerful machines can follow it.

jessesanford:
I like this idea. It really make sense when you look at some of the ephemeral environment tools out there. Uffizzi was one that I had looked at recently. https://github.com/UffizziCloud/uffizzi

PR #505, opened 2025-03-15
docs: add remote host docs

squidboylan:
fixes https://github.com/cnoe-io/idpbuilder/issues/144

### DIFF
--- a/docs/running-idpbuilder-on-a-remote-host.md
+++ b/docs/running-idpbuilder-on-a-remote-host.md
@@ -1,0 +1,49 @@
+# Running IDPBuilder on a remote host
+
+## Option 1: SSH Port forwarding
+
+This option is the most flexible and involves using an ssh connection to forward traffic from local ports to the server where IDPBuilder was run.
+First create your cluster on the server:
+
+```shell
+user@server:~/$ idpbuilder create
+```
+
+Once your cluster is created we need to configure our port forwards:
+
+```shell
+user@local:~/$ ssh -L 8443:server:8443 -L 32222:server:32222 user@server
+```
+
+`-L 8443:server:8443` adds portforwarding for the ingress.
+
+`-L 32222:server:32222` adds portforwarding for the gitea ssh port.
+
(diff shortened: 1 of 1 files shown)


docs/running-idpbuilder-on-a-remote-host.md
```diff
@@ -0,0 +1,49 @@
...
+user@local:~/$ scp user@server:~/.kube/config ~/.kube/config
+```
+
+## Option 2: Changing the ingress host
+
+If you only need remote access to the ingress you can build your remote cluster using the following options:
+
+```shell
+user@server:~/$ idpbuilder create --host SERVER.DOMAIN.NAME.HERE --use-path-routing
+```
```
nabuskey, line 47: I am a bit confused here. Is this for the path routing configuration only? As long as the ingress port is forward to the local machine, the host bit isn't needed? e.g.

```
ssh -L 8443:server:8443
```
This works for use with the `--use-path-routing` flag.

nabuskey (commented):
I know the issue is opened in this repo but the main documentation site is at https://github.com/cnoe-io/website

I'd say it should go to: https://cnoe.io/docs/reference-implementation/installations/idpbuilder/usage

docs/running-idpbuilder-on-a-remote-host.md
squidboylan: I decided to add a second option that avoids the need for port forwarding, that's what this option is. With this option you only change the ingress host and access the ingress over the network rather than a port forward to localhost, path routing is used to avoid the need for  `argocd.SERVER.DOMAIN` and `gitea.SERVER.DOMAIN` dns records. If you think this option is unnecessary I can take it out, I only decided to document it because I had used it in my homelab. Lets continue discussion over on the other repo: https://github.com/cnoe-io/website/pull/101

nabuskey: Ah I see. Yea let's continue this and get it merged in the other repo.

nabuskey:
We will get this done in https://github.com/cnoe-io/website/pull/101

Issue #510, opened 2025-03-31
Feature: Expose the gitea server using NodePort


cmoulliard:
### Have you searched for this feature request?

- [x] I searched but did not find similar requests

### Problem Statement

Until now, a pod running inside another idpbuilder cluster or kind cluster cannot access a gitea server exposed using the ingress url as the hostname is resolved locally against the loopback address.
To access it, we have to expose the gitea server using a NodePort.

### Possible Solution

We can configure the gitea helm chart of idpbuilder to generate the gitea HTTP service as Nodeport using the following values
```
service:
  http:
    type: NodePort
    nodePort: 32223
    externalTrafficPolicy: Local
```
If we do that, then the service will be available using `http://<HOST_IP_ADDRESS>:<NODE_PORT>` or using `https://gitea.cnoes.localtest.me:8443` 

Another change is certainly needed as Argocd reports such an error
```
Failed sync attempt to a7fc045f00389156135e454c93355101a3b94675: one or more objects failed to apply, reason: error when patching "/dev/shm/592575435": Service "my-gitea-http" is invalid: spec.clusterIPs[0]: Invalid value: []string{"None"}: may not change once set (retried 5 times).

// Gitea HTTP Service looks like => 
spec:
  clusterIP: 10.96.128.174
  clusterIPs:
  - 10.96.128.174
  externalTrafficPolicy: Cluster
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - name: http
    nodePort: 32223
    port: 3000
    protocol: TCP
    targetPort: 3000
  selector:
    app.kubernetes.io/instance: my-gitea
    app.kubernetes.io/name: gitea
  sessionAffinity: None
  type: NodePort
```

PR #511, opened 2025-03-31
Generate a NodePort service for the gitea http.

cmoulliard:
-  Generate a NodePort service for the gitea http. 
- See: #510

### DIFF
--- a/hack/gitea/values.yaml
+++ b/hack/gitea/values.yaml
@@ -34,6 +34,11 @@ gitea:
       SKIP_TLS_VERIFY: true
 
 service:
+  http:
+    type: NodePort
+    port: 3000
+    nodePort: 32223
+    externalTrafficPolicy: Local
   ssh:
     type: NodePort
     port: 32222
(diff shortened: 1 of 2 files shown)


punkwalker:
@cmoulliard Is there any specific use-case for which we need this Nodeport? The existing ClusterIP service works fine, right?

cmoulliard:
> The existing ClusterIP service works fine, right?

Yes, it works fine but it is mainly used for internal usage when pod access gitea internally using the kube service address OR when the service should be managed behind an ingress host

cmoulliard:
> Is there any specific use-case for which we need this Nodeport?

As mentioned this is to allow a pod to access gitea from another kind cluster.

cluster 1 : kind cluster running gitea = gitea.cnoe.localtest.me:8443
cluster 2 : kind cluster where a pod (aka a git client) will access gitea running on the cluster 1. In this case, the pod will resolve the address: gitea.cnoe.localtest.me to localhost/127.0.0.01 and will not be able to access: gitea.cnoe.localtest.me:8443

punkwalker (approved):
LGTM 🚀

PR #516, opened 2025-04-25
fix: Fixed a bug where the ArgoCD Kustomization created duplicate argocd-rbac-cm resources

### LINKED ISSUE #515: [Bug]: Duplicate ArgoCD RBAC ConfigMap due to misconfigured Kustomization
### What is your environment, configuration, and command?

In the current setup, when running `idpbuilder create` the ArgoCD manifests generated via `https://raw.githubusercontent.com/argoproj/argo-cd/v2.10.7/manifests/install.yaml` and ` ./hack/argo-cd/generate-manifests.sh` results in the generation of two `argocd-rbac-cm` resources. This is because the Kustomization references the `argocd-rbac-dev.yaml` file as a separate resource rather than a patch causing ArgoCD to have a conflict between those files. 

Image

### What did you do and What did you see instead?

I want idpbuilder to properly patch the existing argocd-rbac-cm-

### Additional Information. Logs.

_No response_

lieberlois:
Resolves #515 by properly patching the existing argocd-rbac-cm.

### DIFF
--- a/hack/argo-cd/argocd-rbac-dev.yaml
+++ b/hack/argo-cd/argocd-rbac-dev.yaml
@@ -1,11 +1,7 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  labels:
-    app.kubernetes.io/name: argocd-rbac-cm
-    app.kubernetes.io/part-of: argocd
   name: argocd-rbac-cm
-  namespace: argocd
 data:
   policy.csv: |
     p, role:developer, applications, *, *, allow
--- a/hack/argo-cd/kustomization.yaml
+++ b/hack/argo-cd/kustomization.yaml
@@ -2,7 +2,6 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
   - https://raw.githubusercontent.com/argoproj/argo-cd/v2.10.7/manifests/install.yaml
-  - argocd-rbac-dev.yaml
 
 patches:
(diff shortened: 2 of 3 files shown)


lieberlois:
@nabuskey Sure, I'll get to it tomorrow!  Just curious whether you think this test is actually useful? Personally I think we are more or less just testing Kustomize here - either way, I'll get to it tomorrow!

nabuskey:
> @nabuskey Sure, I'll get to it tomorrow! Just curious whether you think this test is actually useful? Personally I think we are more or less just testing Kustomize here - either way, I'll get to it tomorrow!

IMO it's not really useful. The test needs to actually validate contents of the manifests. I am ok with removing it. If you do remove it, could you open an issue to track it?

lieberlois:
I'll look into it tomorrow @nabuskey - just wanted to mention my concern here that the test (even if it would validate the contents) doesn't really test idpbuilder but rather ArgoCD Kustomizations or even Kustomize itself. It would be a different story if the test would cover custom overlays (e.g. custom ArgoCD config) applied from idpbuilder - but I'm not even sure if that is possible ^^

lieberlois:
@nabuskey I adjusted the test from 59 to 58 for now since changing the test setup should not be part of this PR. If you want we can discuss this further, but I'd like to get this fixed ASAP since I need it for a workshop 😉

nabuskey (approved):
LGTM

PR #543, opened 2025-08-30
fix: add tests for using the gitea registry in the cluster

squidboylan:


squidboylan:
I confirmed this fails on the 0.10.0 release but works on main

blakeromano (approved):
Looks like a good E2E test case! Thank you!

PR #555, opened 2025-12-18
Add certificate export command for Docker registry trust

greghaynes:
Implements 'idpbuilder get certificate' command to export the self-signed TLS certificate from the cluster. This solves the Docker trust issue with the Gitea container registry without requiring a Docker daemon restart.

Features:
- Default: Prints certificate to stdout for flexibility
- --docker flag: Exports to Docker's per-registry cert directory (~/.docker/certs.d/<registry-host>/ca.crt)
- --output flag: Exports to a custom file path
- Provides platform-specific instructions for system-wide trust
- Works without Docker restart, avoiding cluster shutdown

Includes comprehensive test coverage for certificate retrieval, file operations, registry host determination, and command flags.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ HELM_TGZ ?= $(LOCALBIN)/helm.tar.gz
 HELM ?= $(LOCALBIN)/helm
 
 ## Tool Versions
-CONTROLLER_TOOLS_VERSION ?= v0.15.0
+CONTROLLER_TOOLS_VERSION ?= v0.20.0
 
 .PHONY: fmt
 fmt: ## Run go fmt against code.
--- a/pkg/cmd/get/root.go
+++ b/pkg/cmd/get/root.go
@@ -22,7 +22,7 @@ func init() {
 	GetCmd.AddCommand(ClustersCmd)
 	GetCmd.AddCommand(SecretsCmd)
 	GetCmd.AddCommand(PackagesCmd)
-        GetCmd.AddCommand(CertificateCmd)
+	GetCmd.AddCommand(CertificateCmd)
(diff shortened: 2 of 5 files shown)


greghaynes:
@copilot merge in main

pkg/cmd/get/certificate.go
```diff
@@ -0,0 +1,179 @@
...
+	config, err := util.GetConfig(ctx)
+	if err != nil {
+		return fmt.Errorf("getting idp config: %w", err)
+	}
+
+	var registryHost string
+	if config.UsePathRouting {
+		registryHost = fmt.Sprintf("%s:%s", config.Host, config.Port)
+	} else {
+		registryHost = fmt.Sprintf("gitea.%s:%s", config.Host, config.Port)
```
squidboylan, line 79: I don't think `config.Host` is the right variable to use here, I think `config.IngressHost` is more appropriate. It's not totally clear to me why we have these two config options separated but they can be out of sync and when they are `IngressHost` should be correct I think.

squidboylan:
This feels like something we should have an e2e test for?

pkg/cmd/get/certificate.go
```diff
@@ -0,0 +1,179 @@
...
+	certOutputPath string
+	setupDocker    bool
+)
+
+var CertificateCmd = &cobra.Command{
+	Use:   "certificate",
+	Short: "Export the TLS certificate from the cluster",
+	Long: `Export the self-signed TLS certificate from the cluster.
+
+By default, the certificate is printed to stdout. Use the --docker flag to automatically
```
cmoulliard, line 27: Such a command is specific to docker and will not work for podman certainly. Add then a remark or enhance the code to support to export the certificate for docker or podman

PR #564, opened 2026-02-18
feat: registry mirrors

### LINKED ISSUE #563: Feature: local registry mirrors
### Have you searched for this feature request?

- [x] I searched but did not find similar requests

### Problem Statement

Idpbuilder uses kind which does not use the local image cache of the underlying container engine. Instead each time the IDP is spun up from scratch all images have to be fetched from the internet. This noticeably slows down the creation and causes unnecessary traffic, which is especially apparent while traveling with slow/interrupted connectivity or behind a corporate VPN (or similar).

### Possible Solution

Idpbuilder could allow the user to specify registry mirrors that are then integrated into the containerd node configurations. This way the user is free to spin up lightweight `registry:3` containers with persistent storage:

```bash
idpbuilder \
  --registry-mirrors docker.io=http://kind-registry-docker:5000 \
  --registry-mirrors registry.k8s.io=http://kind-registry-k8s:5000 \
  --registry-mirrors docker.gitea.com=http://kind-registry-gitea:5000
```

The registry mirrors can be managed externally, e.g. by running

```bash
docker volume create kind-registry-docker || true
docker run --rm \
   --name "kind-registry-docker" \
   -e REGISTRY_PROXY_REMOTEURL="https://docker.io" \
   -e OTEL_TRACES_EXPORTER=none \
   -v kind-registry-docker:/var/lib/registry \
   --network kind \
   -d registry:3
```

Unfortunately the `registry` does not support multiple proxy-targets in pull through mode, so one container per mirrored registry is required.

### Alternatives Considered

- vind (vCluster in Docker) uses the local docker image storage out of the box, however implementation is more complex, the feature is fairly new and the only supported container engine is docker.
- management of the registry mirrors could be integrated at the expense of increased complexity and lower configurability of the registries itself, while providing better ergonomics. Additionally, we explicitly do not want to couple the registries lifecycle to the IDP (although the registry could be started/stopped together with the IDP while retaining the volumes).

Argannor:
This MR adds the option to configure registry mirrors.

Closes #563


pkg/kind/resources/hosts-mirror.toml.tmpl
```diff
@@ -0,0 +1,5 @@
+server = "{{ .RegistryAddress }}"
+
+[host."{{ .RegistryAddress }}"]
+  capabilities = ["pull", "resolve"]
+  skip_verify = true
```
squidboylan: it would be nice if we had a way to make this optional, not sure if there's an elegant way to do so though.

squidboylan:
Looks great overall, left one small comment, hoping we can get a second look by someone like @punkwalker

pkg/kind/resources/hosts-mirror.toml.tmpl
Argannor: We can remove it be default and 
1. introduce an extra flag to skip verify for mirrors, or
2. make it part of the current flags syntax 
3. skip verify only for registries configured as http://

squidboylan: we could consider adding a second flag for insecure registries? what do you think about that? Maybe `--insecure-registry-mirror` that follows the same code path but toggles verification skip? That way it can be configured for individual mirrors. What do you think?

Argannor: Added `--insecure-registry-mirrors` (plural, as it applies to all specified mirrors) as an optional flag that defaults to false if not set. The `skip_verify = true` line will be omitted if the flag is not present.

Argannor:
Hi @squidboylan, thank you for the initial review. Code is now adjusted, @punkwalker do you mind taking a look so we can finalize this?

pkg/cmd/create/root.go
```diff
@@ -208,7 +221,17 @@ func validate() error {
...
+		return err
+	}
+
+	// Validate registry mirrors
+	_, err = parseRegistryMirrors(registryMirrors)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
punkwalker: nit: Can we just return the err here.

pkg/cmd/create/root.go
```diff
@@ -240,6 +263,33 @@ func getPackageCustomFile(input string) (v1alpha1.PackageCustomization, error) {
...
+
+		target := strings.TrimSpace(parts[0])
+		address := strings.TrimSpace(parts[1])
+
+		if target == "" {
+			return nil, fmt.Errorf("target registry cannot be empty in mirror: %s", mirror)
+		}
+		if address == "" {
+			return nil, fmt.Errorf("registry address cannot be empty in mirror: %s", mirror)
+		}
```
punkwalker, line 279: should we check for valid URL format for address?

punkwalker (approved):
@Argannor Few comment otherwise LGTM

pkg/cmd/create/root.go
Argannor: fixed

pkg/cmd/create/root.go
Argannor: Added input validation and reference to the containerd docs for both target (real registry) and address of the mirror, and added test cases for the newly added validation

Argannor:
Thank you for the review @punkwalker! I adjusted the code according to your comments and ran `make build` to regenerate the manifests. I was using a >1.25 version of go, which failed the build earlier locally, so I only ran the tests manually before, sorry!

PR #565, opened 2026-03-04
feat: replace ingress-nginx with Traefik v3 as the ingress controller

squidboylan:
I decided to tackle this as a middle ground before tackling migrating to gateway api. Happy to toss this if we want to go straight to gateway API. Interested in hearing what others think

There shouldn't be anything surprising here but there are some implementation details due to traefik's quirks vs nginx including renaming traefik pod port names from web and websecure to http and https, and hardcoding the ingress loadbalancer IP to localhost which prevents the ArgoCD applications from getting stuck in the "Progressing" state.

Additionally in `pkg/k8s/deserialize.go` we add functionality to applying resources that arent part of our schema by representing them as unstructured manifests. This prevents the need to import all of the traefik CRDs into our code which isn't very convenient and these CRs are only applied by reading them off disk, we dont build any in go code.


pkg/build/templates/coredns/cm-coredns-default.yaml.tmpl
```diff
@@ -7,10 +7,10 @@ data:
   default.conf: |
     # Goal: Rewrite rules for in-cluster access to a service: gitea, argocd, etc using the same FQDN as for external access
 
-    # subdomain names e.g. gitea.cnoe.localtest.me resolves to the IP address of the kubernetes ingress service and then will become ingress-nginx-controller.ingress-nginx.svc.cluster.local
+    # subdomain names e.g. gitea.cnoe.localtest.me resolves to the IP address of the kubernetes ingress service and then will become my-traefik.traefik.svc.cluster.local
```
RafPe, line 10: Would it make sense to make it similar to what we have so like maybe omething in line of `ingress-controller.traefik.svc.cluster.local` ?

RafPe (commented):
Looks good from here - only thing I would change is most likely the service name for ingress

RafPe:
Regarding going straight to gateway-api - if thats supported by all the frameworks we plan to use then that also might be idea - as otherwise we will confuse users first with switch to Traefik and then I would guess shortly after we will migrate them to gateway api

squidboylan:
> Regarding going straight to gateway-api - if thats supported by all the frameworks we plan to use then that also might be idea - as otherwise we will confuse users first with switch to Traefik and then I would guess shortly after we will migrate them to gateway api

I like the idea of having an option that is secure and doesnt introduce large breaking changes like gateway api. Additionally I don't have a PR for implementing gateway api at the moment.

